PR #8, opened 2020-06-30
Generate StackOverflow result using hardcoded url

irisliu77:


irisliu77:
@googlebot I signed it!

kaylaleung:
I think the style.css, script.js, and index.html files in this PR can be omitted since Aaron will be working on the hompage/srp?
and queryServlet should be merged into searchServlet.

PR #10, opened 2020-07-01
create simple homepage and results page

AaronLopes:


src/main/webapp/WEB-INF/tags/base.tag
```diff
@@ -1,4 +1,2 @@
 <%@tag description="Common page content" pageEncoding="UTF-8"%>
-<p>
-  This text appears on every page! Put common headers and navigational elements here.
-</p>
+<h3>CodeSnippet</h3>
```
kaylaleung, line 2: Make a separate PR to confirm project name?

src/main/webapp/WEB-INF/templates/homepage.jsp
```diff
@@ -5,15 +5,11 @@
...
   </head>
   <body>
     <t:base/>
-    <p>Hello World!</p>
-    <p>The current date is ${requestScope["date"]}</p>
-    <form method="GET" action="/search">
-      <input name="q" placeholder="Query goes here" required>
-      <input type="submit">
-    </form>
+    <div class="gcse-searchbox-only" data-resultsUrl="/search"></div>
```
kaylaleung, line 13: The homepage placement of the search box is identical to the srp placement. Will this be changed to look closer actually Google homepage with a page centered searchbox?

src/main/webapp/WEB-INF/templates/homepage.jsp
```diff
@@ -5,15 +5,11 @@
 <html>
   <head>
     <t:head/>
+    <link href="../searchbox-style.css" rel="stylesheet" type="text/css">
     <title>Project Homepage</title>
```
irisliu77, line 9: might need to change to our project name

irisliu77: i mean for the title

src/main/webapp/style.css
```diff
@@ -0,0 +1,7 @@
+h3 {
+  text-align: center;
+}
+
+#results-box {
+  width: 50%;
+}
```
kaylaleung, line 7: Will this width styling be compatible with adding a RHS column on the page with a mock card?

src/main/webapp/WEB-INF/templates/homepage.jsp
AaronLopes: For the MVP, I think the unchanging searchbox placement is fine. The placement (along with the searchbox styling) can definitely be changed later to resemble the Google homepage.

src/main/webapp/style.css
AaronLopes: Maybe, but changed to a flexbox bc it'll definitely be easier to work with

src/main/webapp/style.css
```diff
@@ -0,0 +1,8 @@
+h3 {
+  text-align: center;
+}
+
+.srp-container {
+  display: flex;
+  flex-direction: row;
```
anbenson, line 7: This is the default - this is typically left out.

src/main/webapp/style.css
anbenson: It's worth thinking early on about how you want to divide up the width. Is it percentage-based? Hardcoded pixel values? Will there be gaps? Dependent on screen size?

Flexbox definitely works here, but inline-block may be sufficient for your use case as well.

src/main/webapp/WEB-INF/templates/homepage.jsp
anbenson: From the design doc I got the impression that there wasn't going to be a homepage anymore. Has that changed?

Also, the design doc's mock's searchbox doesn't extend across the entire page, but that's not super important to address now.

src/main/webapp/WEB-INF/tags/head.tag
```diff
@@ -1,2 +1,3 @@
 <%@tag description="Common head content" pageEncoding="UTF-8"%>
-<!-- This is where common <link> and <script> tags would go. -->
+<link href="../../style.css" rel="stylesheet" type="text/css">
```
anbenson, line 2: The href refers to the filepath at _runtime_ (since that's when it's evaluated). At runtime, all the current pages are served from / or /search, so the ../.. is unnecessary. To check what I'm saying, try loading localhost:8080/../../style.css - it'll redirect to localhost:8080/style.css.

src/main/webapp/WEB-INF/templates/homepage.jsp
```diff
@@ -5,15 +5,11 @@
 <html>
   <head>
     <t:head/>
-    <title>Project Homepage</title>
+    <link href="../searchbox-style.css" rel="stylesheet" type="text/css">
```
anbenson, line 8: Same thing here - except this file doesn't appear to exist in the PR.

anbenson, line 8: type="text/css" is not necessary in HTML5.

src/main/webapp/WEB-INF/tags/head.tag
```diff
@@ -1,2 +1,3 @@
...
+<script async src="https://cse.google.com/cse.js?cx=017022763452466193978:ejddswiptcs"></script>
```
anbenson, line 3: Is it safe for this script to be async? I assume it's looking for the gcse-searchbox-only class and so forth, which if the script is fetched asynchronously, you could run into a race condition where the rest of the doc isn't parsed yet.

src/main/webapp/style.css
AaronLopes: Considering that we are displaying two main elements (search results and knowledge cards) the space will be shared 50/50. I want to proceed with flexbox for its (relative) simplicity, as I don't see any need to hardcode width values for now.

src/main/webapp/WEB-INF/tags/head.tag
```diff
@@ -1,2 +1,3 @@
 <%@tag description="Common head content" pageEncoding="UTF-8"%>
-<!-- This is where common <link> and <script> tags would go. -->
+<link href="style.css" rel="stylesheet">
+<script src="https://cse.google.com/cse.js?cx=017022763452466193978:ejddswiptcs"></script>
```
anbenson, line 3: With the `async` keyword, my previous comment said there might be a race condition if this script loaded before parsing the rest of the page. Without it, it makes the script synchronously downloaded, making the race condition always occur. But obviously that's not happening, so I took a closer look. It looks like your previous approach with `async` is from the [documentation](https://developers.google.com/custom-search/docs/tutorial/implementingsearchbox), and upon closer inspection, it looks like the script waits for the document to finish loading first, so `async` is fine after all.

Sorry for the back-and-forth.

src/main/webapp/style.css
anbenson: Quick note - style.css is in head.tag, so it'll be included in both the homepage and the srp, but the .srp-container rule is specific to the srp. With the amount of CSS you have currently, it's not going to be an issue to have unused CSS on the homepage, but consider separating out common CSS and page-specific CSS.

PR #12, opened 2020-07-07
Add Card class

kaylaleung:
Add class object with helper functions by Iris for Gson to later convert Card class into JSON object

### DIFF
--- a/src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/data/Card.java
+++ b/src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/data/Card.java
@@ -1,0 +1,54 @@
+package com.google.step.YOUR_PROJECT_NAME_HERE.data;
+
+public final class Card {
+  private String title;
+  private String code;
+  private String link;
+  private String description;
+
+  public Card() {
+    this.title = "";
+    this.code = "";
+    this.link = "";
+    this.description = "";
+  }
+
+  public Card(String title, String code, String link, String description) {
+    this.title = title;
+    this.code = code;
+    this.link = link;
+    this.description = description;
+  }
+
+  public void setTitle(String title) {
+    this.title = title;
+  }
+
+  public String getTitle() {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/data/Card.java
```diff
@@ -0,0 +1,54 @@
...
+  private String code;
+  private String link;
+  private String description;
+
+  public Card() {
+    this.title = "";
+    this.code = "";
+    this.link = "";
+    this.description = "";
+  }
```
irisliu77, line 14: this looks strange to me. i think it should be an default constructor

kaylaleung: ah, good point, I set the variables to initialize as empty instead, the default constructor shouldn't need to be explicitly generated

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/data/Card.java
```diff
@@ -6,13 +6,6 @@
   private String link;
   private String description;
```
irisliu77, line 8: since we specified a constructor, the default constructor will not automatically occur here. will need to have something like this 
```
public Card() {}
```

irisliu77: at least for my code, an empty constructor is needed since i cannot get all field of the result card in the same method. So being able to make an empty instance would help.

anbenson: My understanding of the current structure of the code is that the Card class is just a container around these 4 fields, which should have no reason to change once they've been set. So I strongly recommend that this class be made immutable e.g. the fields are set in the constructor, which means there'd be no need for a default constructor or any setters.

PR #13, opened 2020-07-07
Add Stackoverflow client

irisliu77:
Add stackoverflow client that can create a card from stackoverflow search result


src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,157 @@
...
+      "https://api.stackexchange.com/2.2/questions/%s/answers?"
+          + "order=desc&sort=votes&site=stackoverflow";
+  private static final String ANSWER_URL_TEMPLATE =
+      "https://api.stackexchange.com/2.2/answers/%s?order"
+          + "=desc&sort=activity&site=stackoverflow&filter=!9_bDE(fI5";
+
+  public Card search(String so_url) {
+    Card card = getQuestion(so_url);
+    card = getAnswerId(card);
+    card = getAnswer(card);
```
kaylaleung, line 36: Instead of updating and returning an updated card with each of these functions, can this instead just return the piece of card information that each function is updating so that after gathering all the 4 pieces of information required to create a card, we can check of all the info is available, and then create a card at the very end?

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,157 @@
...
+    if (so_url == null) {
+      return card;
+    }
+    CloseableHttpClient httpClient = HttpClients.createDefault();
+    try {
+      URI uri = new URI(so_url);
+      System.out.println("path = " + uri.getPath());
+      // Parse the URL to get the question id.
+      String[] segments = uri.getPath().split("/");
+      String idStr = segments[2];
```
kaylaleung, line 52: Magic numbers here should be avoided or declared as constant with comment?

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,157 @@
...
+        responseBody.append(line);
+      }
+      JSONObject res = new JSONObject(responseBody.toString());
+      String title = res.getJSONArray("items").getJSONObject(0).get("title").toString();
+      card.setLink(uri.toString());
+      // Store the id of the question in order to get the code body of the answer.
+      card.setCode(idStr);
+      card.setTitle(title);
+    } catch (Exception e) {
+      e.printStackTrace();
```
kaylaleung, line 75: This is a really long try/catch block. Maybe have the function throw an exception to be caught by the calling function instead of catching the exception here?

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,157 @@
...
+      String line = null;
+      while ((line = reader.readLine()) != null) {
+        responseBody.append(line);
+      }
+      JSONObject res = new JSONObject(responseBody.toString());
+      String title = res.getJSONArray("items").getJSONObject(0).get("title").toString();
+      card.setLink(uri.toString());
+      // Store the id of the question in order to get the code body of the answer.
+      card.setCode(idStr);
+      card.setTitle(title);
```
kaylaleung, line 73: Is there a way to return this information distinctly, rather than setting it here directly as the fields of the card such that the card is only created when all pieces of information exist? i.e. have different helper functions, each responsible for returning just the code, title, desc, link etc?

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,157 @@
...
+      while ((line = reader.readLine()) != null) {
+        responseBody.append(line);
+      }
+      JSONObject res = new JSONObject(responseBody.toString());
+      JSONArray answer_list = res.getJSONArray("items");
+      String answer_id = res.getJSONArray("items").getJSONObject(0).get("answer_id").toString();
+      // Replace the question id by the answer id in order to retrieve the code body next.
+      card.setCode(answer_id);
+    } catch (ClientProtocolException e) {
+      e.printStackTrace();
```
kaylaleung, line 105: Long try/catch block here as well

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,157 @@
...
+      description = description.substring(0, 99);
+      // Combine all code in the answer body.
+      Elements code_html = doc.select("code");
+      String code = "";
+      for (Element e : code_html) {
+        code += e.outerHtml();
+      }
+      card.setDescription(description);
+      card.setCode(code);
+    } catch (ClientProtocolException e) {
```
kaylaleung, line 150: Long try/catch block here as well

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/data/Card.java
```diff
@@ -0,0 +1,49 @@
...
+  public String getLink() {
+    return this.link;
+  }
+
+  public void setDescription(String description) {
+    this.description = description;
+  }
+
+  public String getDescription() {
+    return this.description;
```
kaylaleung, line 47: If you can return the values rather than set them in the helper functions themselves, we can actually probably avoid having to use these helper functions.

kaylaleung: Because also, I don't think getDescription/getLink/getTitle are used anywhere else in your code

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,157 @@
...
+public final class StackOverflowClient {
+  private static final String SEARCH_URL_TEMPLATE =
+      "https://api.stackexchange.com/2.2/questions/%s?"
+          + "order=desc&sort=activity&site=stackoverflow";
+  private static final String QUESTION_URL_TEMPLATE =
+      "https://api.stackexchange.com/2.2/questions/%s/answers?"
+          + "order=desc&sort=votes&site=stackoverflow";
+  private static final String ANSWER_URL_TEMPLATE =
+      "https://api.stackexchange.com/2.2/answers/%s?order"
+          + "=desc&sort=activity&site=stackoverflow&filter=!9_bDE(fI5";
```
kaylaleung, line 31: Is !9_bDE(fI5 here supposed to be a specific filter type? Unclear what it's filtering for.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,157 @@
...
+      if (entity == null) {
+        return card;
+      }
+      BufferedReader reader = new BufferedReader(new InputStreamReader(entity.getContent()));
+      StringBuilder responseBody = new StringBuilder();
+      String line = null;
+      while ((line = reader.readLine()) != null) {
+        responseBody.append(line);
+      }
+      JSONObject res = new JSONObject(responseBody.toString());
```
kaylaleung, line 131: Most of the code here is repeated three times across all three helper functions. Is there a way you can minimize the repetition by moving this code into a separate helper function?

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/data/Card.java
irisliu77: need to use getCode() to store the question_id, answer_id before actually being able to get the code body

kaylaleung: Can you pass in the question_id, answer_id as arguments for your function that retrieves the code body rather than storing them in the code field of the card via getCode()?

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,129 @@
...
+  private static final int ID_INDEX = 2;
+
+  public Card search(String so_url) {
+    try {
+      Card card = getQuestion(so_url);
+      card = getAnswerId(card);
+      card = getAnswer(card);
+      return card;
+    } catch (URISyntaxException e) {
+      return new Card();
```
kaylaleung, line 41: I think this function should just return null if a proper card cannot be constructed or if not all fields of the Card can be populated, rather than returning an empty new card. The null Card check can happen in SearchServlet

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  /* Get the question id based on url from the CSE result */
+  private Card getQuestion(String so_url) throws URISyntaxException {
```
kaylaleung, line 46: Avoid underscore parameter naming (I think?)

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,129 @@
...
+    return card;
+  }
+
+  /* Get the most voted answer's id and store it in the card */
+  private Card getAnswerId(Card card) {
+    String url = String.format(QUESTION_URL_TEMPLATE, card.getCode());
+    try {
+      JSONObject res = getResponse(url);
+      JSONArray answer_list = res.getJSONArray("items");
+      String answer_id = res.getJSONArray("items").getJSONObject(0).get("answer_id").toString();
```
kaylaleung, line 72: Avoid nonconstant underscore variable naming here and elsewhere in the file and use camel case instead (?)

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,129 @@
...
+      String body = res.getJSONArray("items").getJSONObject(0).get("body").toString();
+      Document doc = Jsoup.parse(body);
+      // Combine all description in the answer body.
+      Elements description_html = doc.select("p");
+      String description = "";
+      for (Element e : description_html) {
+        description += e.outerHtml();
+      }
+      // Get the first 100 characters for displaying.
+      description = description.substring(0, 99);
```
kaylaleung, line 95: Also set this as a number constant and move comment up with it?

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
irisliu77: note: also change the items to constant

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,136 @@
...
+      String body =
+          res.getJSONArray(ITEM_PARAMETER).getJSONObject(0).get(BODY_PARAMETER).toString();
+      Document doc = Jsoup.parse(body);
+      // Combine all description in the answer body.
+      Elements descriptionHtml = doc.select("p");
+      String description = "";
+      for (Element e : descriptionHtml) {
+        description += e.outerHtml();
+      }
+      description = description.substring(0, DESCRIPTION_LENGTH_PARAMETER);
```
kaylaleung, line 103: This raises n java.lang.StringIndexOutOfBoundsException error when the description parameter is less than 200. Should add a condition where if the description is less than 200, it just sets the whole thing to the desc

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,134 @@
...
+          + "=desc&sort=activity&site=stackoverflow&filter=!9_bDE(fI5";
+  private static final int ID_INDEX = 2;
+  private static final String ITEM_PARAMETER = "items";
+  private static final String TITLE_PARAMETER = "title";
+  private static final String BODY_PARAMETER = "body";
+  private static final String CODE_PARAMETER = "code";
+  private static final String ANSWER_ID_PARAMETER = "answer_id";
+  private static final int DESCRIPTION_LENGTH_PARAMETER = 200;
+
+  public Card search(String stackLink) {
```
joanchenx, line 38: Nit: Consider renaming `stackLink` to `url` here and elsewhere since your class is called StackOverlfowClient, it's clear that this should be a StackOverflow link.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+
+  /* Get the question id based on url from the CSE result */
+  private Card getQuestion(String stackLink) throws URISyntaxException {
+    Card card = new Card();
+    URI uri = new URI(stackLink);
+    // Parse the URL to get the question id.
+    String[] segments = uri.getPath().split("/");
+    String idStr = segments[ID_INDEX];
```
joanchenx, line 56: Nit: rename `idStr` to `questionId` since you use `answerId` in the next method and I think that helps make it more clear.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,134 @@
...
+import org.jsoup.select.Elements;
+
+public final class StackOverflowClient {
+  private static final String SEARCH_URL_TEMPLATE =
+      "https://api.stackexchange.com/2.2/questions/%s?"
+          + "order=desc&sort=activity&site=stackoverflow";
+  private static final String QUESTION_URL_TEMPLATE =
+      "https://api.stackexchange.com/2.2/questions/%s/answers?"
+          + "order=desc&sort=votes&site=stackoverflow";
+  // this url specify filter to generate answer body
```
joanchenx, line 26: Nit: `// This URL specified a filter to generate the answer body.`

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,134 @@
...
+      card = getAnswerId(card);
+      card = getAnswer(card);
+      return card;
+    } catch (URISyntaxException e) {
+      // return null card if no valid card available
+      return null;
+    }
+  }
+
+  /* Get the question id based on url from the CSE result */
```
joanchenx, line 50: Nit: capitalize `URL` and add a period at the end of the sentence.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,134 @@
...
+  private Card getQuestion(String stackLink) throws URISyntaxException {
+    Card card = new Card();
+    URI uri = new URI(stackLink);
+    // Parse the URL to get the question id.
+    String[] segments = uri.getPath().split("/");
+    String idStr = segments[ID_INDEX];
+    String url = String.format(SEARCH_URL_TEMPLATE, idStr);
```
joanchenx, line 57: Nit: `searchUrl`

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,134 @@
...
+      card.setTitle(title);
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+    return card;
+  }
+
+  /* Get the most voted answer's id and store it in the card */
+  private Card getAnswerId(Card card) {
+    String url = String.format(QUESTION_URL_TEMPLATE, card.getCode());
```
joanchenx, line 74: Nit: `questionUrl`

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,134 @@
...
+      card.setCode(answerId);
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+    return card;
+  }
+
+  /* Get the content of the answer and store it in the card */
+  private Card getAnswer(Card card) {
+    String url = String.format(ANSWER_URL_TEMPLATE, card.getCode());
```
joanchenx, line 89: Nit: `answerUrl`

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,134 @@
...
+
+  /* Get the most voted answer's id and store it in the card */
+  private Card getAnswerId(Card card) {
+    String url = String.format(QUESTION_URL_TEMPLATE, card.getCode());
+    try {
+      JSONObject res = getResponse(url);
+      String answerId =
+          res.getJSONArray(ITEM_PARAMETER).getJSONObject(0).get(ANSWER_ID_PARAMETER).toString();
+      // Replace the question id by the answer id in order to retrieve the code body next.
+      card.setCode(answerId);
```
joanchenx, line 80: I think it's a bit weird that card.code is being used to store the question and answer ID. It seems like the IDs should be in a separate variable.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
irisliu77: `url` is used in this method and i think `stackLink` is consistent with the servlet

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
kaylaleung: Just tested it now, still think there is an issue here with out of bounds when the description is less than 200? This is the example query/link I tested with [https://stackoverflow.com/questions/8938235/sort-an-array-in-java](https://stackoverflow.com/questions/8938235/sort-an-array-in-java)

webapp (2).iml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kaylaleung, line 1: This file should be omitted?

irisliu77: done, kind of messed up the maven config on my ide lol

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
irisliu77: done, the commit was overwritten by my upstream operation

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,143 @@
...
+    CloseableHttpClient httpClient = HttpClients.createDefault();
+    CloseableHttpResponse response = httpClient.execute(new HttpGet(url));
+    if (response.getStatusLine().getStatusCode() != 200) {
+      return new JSONObject();
+    }
+    HttpEntity entity = response.getEntity();
+    if (entity == null) {
+      return new JSONObject();
+    }
+    BufferedReader reader = new BufferedReader(new InputStreamReader(entity.getContent()));
```
anbenson, line 135: I'm no expert in Java. But I think you need to be closing resources after you use them.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,143 @@
...
+    String[] segments = uri.getPath().split("/");
+    String questionId = segments[ID_INDEX];
+    if (!Pattern.matches("[0-9]+", questionId)) {
+      return null;
+    }
+    String searchUrl = String.format(SEARCH_URL_TEMPLATE, questionId);
+    try {
+      JSONObject res = getResponse(searchUrl);
+      String title =
+          res.getJSONArray(ITEM_PARAMETER).getJSONObject(0).get(TITLE_PARAMETER).toString();
```
anbenson, line 68: Any of these could fail, either by throwing an exception or return null, especially since getResponse's behavior on error is to return an empty JSONObject.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,143 @@
...
+      card.setLink(uri.toString());
+      // Store the id of the question in order to get the code body of the answer.
+      card.setCode(questionId);
```
anbenson, line 71: This is violating your data structure invariant, to put non-code in the code field.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
anbenson: +1, except I think the only line that can throw an error is getResponse...?

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
anbenson: I know I said it was fine yesterday to pass partially-filled Cards around, but this is not what I meant. I would much rather the function return multiple values, as inelegant as it is, than reuse fields for things not advertised as such.

The core problem is that this class is not broken down into tasks in ways where the variables each task needs are in appropriate scopes. For example, here you're being forced to pass questionid and answerid within the Card to make it available to your other methods. Here's a different task breakdown that I believe solves the issue:
search() does the following:
1) get question id (in a private method if you want)
2) pass question id to a private method, get back the title
3) pass question id to a private method, get back the first answer id
4) pass answer id to a private method, get back answer body (I think this is HTML based on your code?)
5) pass answer body to a private method, get back description
6) pass answer body to a private method, get back code
7) finally create your Card object using title, uri, description, code

Thoughts?

(Note how each task either computes one piece of the Card, or does a network request to get a required parameter for a following task.)

(Side-effect: Card can be immutable after all, and all the getters and setters can be deleted.)

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,143 @@
...
+      String body =
+          res.getJSONArray(ITEM_PARAMETER).getJSONObject(0).get(BODY_PARAMETER).toString();
+      Document doc = Jsoup.parse(body);
+      // Combine all description in the answer body.
+      Elements descriptionHtml = doc.select("p");
+      String description = "";
+      for (Element e : descriptionHtml) {
+        description += e.outerHtml();
+      }
+      if (description.length() >= 200) {
```
anbenson, line 108: I'd like to see a comment or something so 200 isn't too magic.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
anbenson: To clarify: I'm okay with passing around partially-filled Cards if it makes it clearer how the code works (e.g. the paradigm is that each method fills out one or two fields of the Card, and this flow is clearly documented). I'm not okay with using the Card as hidden state - here, Card.code becomes an extra hidden argument to every method, and its meaning changes per method, which is _not_ clear at all.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,153 @@
...
+    } catch (URISyntaxException e) {
+      // Return null card if no valid card available.
+      return null;
+    }
+    String title = getTitle(questionId);
+    String answerId = getAnswerId(questionId);
+    // If there is no answer to the question, return a null card.
+    if (answerId == null) {
+      return null;
+    }
```
joanchenx, line 55: Move this before trying to get the title? That way, you don't have to attempt to get the title if the answerId is null.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
joanchenx: Is this a custom filter you've made from https://api.stackexchange.com/docs/filters?

If so, the comment could be something `// This URL specifies a custom StackExchange API filter that retrieves the answer body.`

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,152 @@
...
+    String answerId = getAnswerId(questionId);
+    if (answerId == null) {
+      return null;
+    }
+    String title = getTitle(questionId);
+    String answerBody = getAnswerBody(answerId);
+    String description = getDescription(answerBody);
+    String code = getCode(answerBody);
+    Card card = new Card(title, code, url, description);
+    return card;
```
joanchenx, line 59: Nit: just `return new Card(title, code, url, description);`

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,152 @@
...
+      return null;
+    }
+    return questionId;
+  }
+
+  /* Get the question title using question id */
+  private String getTitle(String questionId) {
+    String searchUrl = String.format(SEARCH_URL_TEMPLATE, questionId);
+    String title = getResponse(searchUrl, TITLE_PARAMETER);
+    return title;
```
joanchenx, line 78: Nit: just `return getResponse(searchUrl, TITLE_PARAMETER);`

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /* Get the most voted answer's id and store it in the card. */
+  private String getAnswerId(String questionId) {
+    String questionUrl = String.format(QUESTION_URL_TEMPLATE, questionId);
+    String answerId = getResponse(questionUrl, ANSWER_ID_PARAMETER);
+    // Replace the question id by the answer id in order to retrieve the code body next.
+    return answerId;
```
joanchenx, line 86: Nit: same here with the `return getResponse(...);` Also, remove the comment as it's no longer relevant.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /* Get the content of the answer and store it in the card. */
+  private String getAnswerBody(String answerId) {
+    String answerUrl = String.format(ANSWER_URL_TEMPLATE, answerId);
+    String body = getResponse(answerUrl, BODY_PARAMETER);
+    return body;
```
joanchenx, line 93: Nit: same here `return getResponse`

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /* Get the question title using question id */
+  private String getTitle(String questionId) {
+    String searchUrl = String.format(SEARCH_URL_TEMPLATE, questionId);
+    String title = getResponse(searchUrl, TITLE_PARAMETER);
+    return title;
+  }
+
+  /* Get the most voted answer's id and store it in the card. */
```
joanchenx, line 81: Since you're not storing the answerId in the card anymore, you can change this to just be `Return the ID of the highest voted answer.`

Likewise, modify the comments for your other `get` methods.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /* Get the content of the answer and store it in the card. */
+  private String getAnswerBody(String answerId) {
+    String answerUrl = String.format(ANSWER_URL_TEMPLATE, answerId);
+    String body = getResponse(answerUrl, BODY_PARAMETER);
+    return body;
+  }
+
+  /* Get the description using return answer body. */
```
joanchenx, line 96: Nit: instead of `return answer body` maybe change to `Return the description parsed from the answer body.`

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/StackOverflowClient.java
```diff
@@ -0,0 +1,152 @@
...
+    for (Element e : descriptionHtml) {
+      description += e.outerHtml();
+    }
+    if (description.length() >= DESCRIPTION_LENGTH_PARAMETER) {
+      description = description.substring(0, DESCRIPTION_LENGTH_PARAMETER);
+    }
+    return description;
+  }
+
+  /* Get the code using return answer body. */
```
joanchenx, line 111: Nit: same here, `Return the code parsed from the answer body.`

pom.xml
```diff
@@ -34,6 +34,21 @@
       <artifactId>json</artifactId>
       <version>20200518</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
```
kaylaleung, line 38: GSON isn't used anywhere in this PR, so just remove this dependency for now

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+    String answerId = getAnswerId(questionId);
+    if (answerId == null) {
+      return null;
+    }
+    String title = getTitle(questionId);
+    String answerBody = getAnswerBody(answerId);
+    String description = getDescription(answerBody);
+    String code = getCode(answerBody);
+    return new Card(title, code, url, description);
```
anbenson, line 60: In Kayla's W3Schools parser, Cards were only returned if all fields are non-empty. The same check isn't done here - is that intentional?

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -0,0 +1,149 @@
...
+    } catch (URISyntaxException e) {
+      // Return null card if no valid card available.
+      return null;
+    }
+    String answerId = getAnswerId(questionId);
+    if (answerId == null) {
+      return null;
+    }
+    String title = getTitle(questionId);
+    String answerBody = getAnswerBody(answerId);
```
anbenson, line 57: This could be null.

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -0,0 +1,149 @@
...
+  /* Return the description parsed from answer body. */
+  private String getDescription(String body) {
+    Document doc = Jsoup.parse(body);
+    // Combine all description in the answer body.
+    Elements descriptionHtml = doc.select("p");
+    String description = "";
+    for (Element e : descriptionHtml) {
+      description += e.outerHtml();
+    }
+    if (description.length() >= DESCRIPTION_LENGTH_PARAMETER) {
```
anbenson, line 102: Consider moving this into the for-loop so you can break early once you've reached the max description length.

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
irisliu77: yes, since stackoverflow sometimes has answer that is without code snippet/description, so the same check will not happen here

PR #14, opened 2020-07-08
Add W3School client to scrape content from URL

kaylaleung:
Retrieve content for card fields: title, link, desc, and code from W3School link. If multiple matches for a scraped tag, just use the first item.


kaylaleung:
Ignore Card class for now. Should be resolved in separate PR. Couldn't omit it here through because maven build check wouldn't pass without Card class

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,43 @@
...
+    Elements descriptions = doc.getElementsByTag(DESC_ID);
+    Elements snippets = doc.getElementsByClass(SNIPPET_ID);
+
+    String title = "";
+    String description = "";
+    String code = "";
+
+    if (titles != null
+        && !titles.isEmpty()
+        && descriptions != null
```
irisliu77, line 28: i would suggest doing null/empty check for each field separately.
```
if (titles == null || titles.isEmpty()) {
  return;
```

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,43 @@
...
+        && snippets != null
+        && !snippets.isEmpty()) {
+      title = titles.first().text();
+      description = descriptions.first().text();
+      code = snippets.first().getElementsByClass(CODE_ID).text();
+    }
+
+    if (!title.isEmpty() && !description.isEmpty() && !code.isEmpty()) {
+      Card w3Card = new Card(title, code, w3Link, description);
+      return w3Card;
```
irisliu77, line 39: what if you cannot retrieve some of the field? i think the returned card can have some empty field.

anbenson:
Drive-by comment: prefer to PR from branches in this repo instead of your own fork so it's easier for everyone to rebase work on top of yours

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
kaylaleung: I don't see any value in returning / displaying half empty cards. I think we already identified 4 essential properties of a Card and I think returning only title/link or returning only code with no title would be more confusing and seemingly incomplete than helpful

irisliu77: i'm thinking of some page that might not have code in it?

kaylaleung: *resolved via discussion*

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
kaylaleung: *resolved via discussion*

pom.xml
```diff
@@ -34,6 +34,16 @@
       <artifactId>json</artifactId>
       <version>20200518</version>
     </dependency>
+    <dependency>
+      <groupId>org.jsoup</groupId>
+      <artifactId>jsoup</artifactId>
+      <version>1.11.3</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
```
kaylaleung, line 43: GSON technically isn't being used in this PR, but it'll end up being used in SearchServlet anyways

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,43 @@
...
+    String title = "";
+    String description = "";
+    String code = "";
+
+    if (titles != null
+        && !titles.isEmpty()
+        && descriptions != null
+        && !descriptions.isEmpty()
+        && snippets != null
+        && !snippets.isEmpty()) {
```
joanchenx, line 31: Since you do this null or empty check quite often, maybe having a helper method would make it more readable?

```
private boolean isNullOrEmpty(Elements elements) {
  return elements == null || elements.isEmpty());
}
```

and then you can do
`if (!isNullOrEmpty(title) && !isNullOrEmpty(descriptions) && !isNullOrEmpty(snippets))`

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,43 @@
+package com.google.step.YOUR_PROJECT_NAME_HERE.external;
+
+import com.google.step.YOUR_PROJECT_NAME_HERE.data.Card;
+import java.io.IOException;
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.jsoup.select.Elements;
+
+public final class W3SchoolClient {
+  private static final String TITLE_ID = "h1";
```
AaronLopes, line 10: nit: change to CARD_TITLE_ID, etc.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,42 @@
...
+
+public final class W3SchoolClient {
+  private static final String CARD_TITLE_ID = "h1";
+  private static final String CARD_DESC_ID = "p";
+  private static final String CARD_SNIPPET_ID = "w3-example";
+  private static final String CARD_CODE_ID = "w3-code";
+
+  public Card search(String w3Link) throws IOException {
+
+    Document doc = Jsoup.connect(w3Link).get();
```
anbenson, line 17: Wow, that's the most succinct GET request I've seen in Java ever.

anbenson: But you could probably make it even more concise with Jsoup.parse? Only necessary if you all end up making the interface to the parsers a URI instead of a String.

anbenson, line 17: I'm guessing this is the part that can throw an IOException. But I don't think it's the caller's job to deal with the search() method not returning a Card. Maybe just return null instead, like you're already doing in the case of other issues?

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    if (!title.isEmpty() && !description.isEmpty() && !code.isEmpty()) {
+      Card w3Card = new Card(title, code, w3Link, description);
+      return w3Card;
+    }
+    return null;
+  }
+
+  public boolean isNullOrEmpty(Elements elements) {
```
anbenson, line 39: Doesn't need to be public. Can be static.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,42 @@
...
+    String code = "";
+
+    if (!isNullOrEmpty(titles) && !isNullOrEmpty(descriptions) && !isNullOrEmpty(snippets)) {
+      title = titles.first().text();
+      description = descriptions.first().text();
+      code = snippets.first().getElementsByClass(CARD_CODE_ID).text();
+    }
+
+    if (!title.isEmpty() && !description.isEmpty() && !code.isEmpty()) {
+      Card w3Card = new Card(title, code, w3Link, description);
```
anbenson, line 33: just `return new Card(...)`?

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,42 @@
+package com.google.step.YOUR_PROJECT_NAME_HERE.external;
```
anbenson, line 1: I didn't actually test this so I'm going to trust you that this works haha :)

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,42 @@
...
+    Document doc = Jsoup.connect(w3Link).get();
+    Elements titles = doc.getElementsByTag(CARD_TITLE_ID);
+    Elements descriptions = doc.getElementsByTag(CARD_DESC_ID);
+    Elements snippets = doc.getElementsByClass(CARD_SNIPPET_ID);
+
+    String title = "";
+    String description = "";
+    String code = "";
+
+    if (!isNullOrEmpty(titles) && !isNullOrEmpty(descriptions) && !isNullOrEmpty(snippets)) {
```
anbenson, line 26: It really feels like you're trying to avoid doing work unless it's going to be needed later on here.

Completely optional, but have you considered just computing one field at a time, and returning `null` early if it turned out to be null/empty/empty string? e.g.

```java
Element titleElement = doc.getElementsByTag(CARD_TITLE_ID).first();
if (titleElement == null || titleElement.text().isEmpty()) {
  return null;
}
String title = titleElement.text();

// and so forth for the rest, and if you didn't return by the end, return a new card
```

anbenson, line 26: getElementsByTag doesn't appear to return null, but just an empty collection. So I don't think this method needs to exist?

pom.xml
```diff
@@ -34,6 +34,16 @@
       <artifactId>json</artifactId>
       <version>20200518</version>
     </dependency>
+    <dependency>
+      <groupId>org.jsoup</groupId>
+      <artifactId>jsoup</artifactId>
+      <version>1.11.3</version>
```
anbenson, line 40: Looks like an old version of jsoup (or at least Iris' PR has a more recent version).

pom.xml
anbenson: None of the active PRs actually use it still...so I would defer until you actually do. No need to bundle this change with an unrelated PR.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.google.step.YOUR_PROJECT_NAME_HERE.data.Card;
+import java.io.IOException;
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.jsoup.select.Elements;
+
+public final class W3SchoolClient {
+  private static final String CARD_TITLE_ID = "h1";
+  private static final String CARD_DESC_ID = "p";
```
anbenson, line 11: These are tags, not IDs.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,42 @@
...
+  private static final String CARD_SNIPPET_ID = "w3-example";
+  private static final String CARD_CODE_ID = "w3-code";
```
anbenson, line 13: These are classes, not IDs.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public Card search(String w3Link) {
+    Document doc;
+    try {
```
anbenson, line 17: try-catch only the part that throws an exception.

src/main/java/com/google/step/YOUR_PROJECT_NAME_HERE/external/W3SchoolClient.java
```diff
@@ -0,0 +1,42 @@
...
+      doc = Jsoup.connect(w3Link).get();
+      Elements titles = doc.getElementsByTag(CARD_TITLE_ID);
+      Elements descriptions = doc.getElementsByTag(CARD_DESC_ID);
+      Elements snippets = doc.getElementsByClass(CARD_SNIPPET_ID);
```
anbenson, line 21: So my last comment about only doing work if you need to - here you still parse out all three of these things before checking if they're empty or whatnot.

PR #15, opened 2020-07-09
Redirect to search.jsp

AaronLopes:
(was previously printing links and not doing this)


kaylaleung:
lgtm, the only thing is that unrelated to this PR, there is a miscellaneous `<div>test</div> `on line 16 of search.jsp that I think was merged onto master

PR #16, opened 2020-07-10
Change project name

kaylaleung:
Change project name to "snippet"

Project ID still set to "step1-2020" because I don't think I have permissions to change the cloud console project name that we deploy to?


irisliu77:
lgtm!

anbenson:
Yeah, I don't think you can change the deploy ID, so I chose step1-2020 instead of YOUR_PROJECT_ID_HERE :)

PR #18, opened 2020-07-10
Revert http code to be compatible with Java 8 deploy

kaylaleung:
Refactor servlet code to use Apache HttpComponent for it to be deployable with Java 8 app engine

### DIFF
--- a/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
+++ b/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
@@ -14,10 +14,9 @@
 
 package com.google.step.snippet.servlets;
 
-import com.google.gson.JsonIOException;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
 import com.google.gson.JsonParser;
-import com.google.gson.JsonSyntaxException;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.UnsupportedEncodingException;
@@ -109,7 +108,6 @@ public class SearchServlet extends HttpServlet {
     String url = CSE_URL + "?key=" + API_KEY + "&cx=" + id + "&q=" + query;
     try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
       try (CloseableHttpResponse response = httpClient.execute(new HttpGet(url))) {
-
         if (response.getStatusLine().getStatusCode() != 200) {
           return null;
         }
@@ -117,24 +115,24 @@ public class SearchServlet extends HttpServlet {
         if (entity == null) {
           return null;
         }
-        JsonParser parser = new JsonParser();
         JsonObject obj;
         try {
+          JsonParser parser = new JsonParser();
           obj = parser.parse(new InputStreamReader(entity.getContent())).getAsJsonObject();
-        } catch (JsonIOException
-            | JsonSyntaxException
+        } catch (JsonParseException
+            | IllegalStateException
             | UnsupportedOperationException
             | IOException e) {
           return null;
         }
-        if (!obj.isJsonNull()
-            && !obj.get(CSE_ITEMS).getAsJsonArray().isJsonNull()
-            && obj.get(CSE_ITEMS).getAsJsonArray().size() > 0) {
-          JsonObject item = obj.get(CSE_ITEMS).getAsJsonArray().get(0).getAsJsonObject();
-          if (!item.isJsonNull() && !item.get(CSE_LINK).isJsonNull()) {
-            return item.get(CSE_LINK).getAsString();
+        if (obj.has(CSE_ITEMS) && obj.getAsJsonArray(CSE_ITEMS).size() > 0) {
+          JsonObject topResult = obj.getAsJsonArray(CSE_ITEMS).get(0).getAsJsonObject();
+          if (topResult.has(CSE_LINK)) {
+            return topResult.get(CSE_LINK).getAsString();
           }
         }
+      } catch (IOException | IllegalStateException | ClassCastException e) {
+        return null;
       }
     } catch (IOException e) {
       return null;
--- a/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
+++ b/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
@@ -107,37 +107,37 @@ public class SearchServlet extends HttpServlet {
 
   private String getLink(String id, String query) {
     String url = CSE_URL + "?key=" + API_KEY + "&cx=" + id + "&q=" + query;
-    CloseableHttpClient httpClient = HttpClients.createDefault();
-    CloseableHttpResponse response;
-    try {
-      response = httpClient.execute(new HttpGet(url));
+    try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
+      try (CloseableHttpResponse response = httpClient.execute(new HttpGet(url))) {
+
+        if (response.getStatusLine().getStatusCode() != 200) {
+          return null;
+        }
+        HttpEntity entity = response.getEntity();
+        if (entity == null) {
+          return null;
+        }
+        JsonParser parser = new JsonParser();
+        JsonObject obj;
+        try {
+          obj = parser.parse(new InputStreamReader(entity.getContent())).getAsJsonObject();
+        } catch (JsonIOException
+            | JsonSyntaxException
+            | UnsupportedOperationException
+            | IOException e) {
+          return null;
+        }
+        if (!obj.isJsonNull()
+            && !obj.get(CSE_ITEMS).getAsJsonArray().isJsonNull()
+            && obj.get(CSE_ITEMS).getAsJsonArray().size() > 0) {
+          JsonObject item = obj.get(CSE_ITEMS).getAsJsonArray().get(0).getAsJsonObject();
+          if (!item.isJsonNull() && !item.get(CSE_LINK).isJsonNull()) {
+            return item.get(CSE_LINK).getAsString();
+          }
+        }
+      }
     } catch (IOException e) {
       return null;
-    }
-    if (response.getStatusLine().getStatusCode() != 200) {
-      return null;
-    }
-    HttpEntity entity = response.getEntity();
-    if (entity == null) {
-      return null;
-    }
-    JsonParser parser = new JsonParser();
-    JsonObject obj;
-    try {
-      obj = parser.parse(new InputStreamReader(entity.getContent())).getAsJsonObject();
-    } catch (JsonIOException
-        | JsonSyntaxException
-        | UnsupportedOperationException
-        | IOException e) {
-      return null;
-    }
-    if (!obj.isJsonNull()
-        && !obj.get(CSE_ITEMS).getAsJsonArray().isJsonNull()
-        && obj.get(CSE_ITEMS).getAsJsonArray().size() > 0) {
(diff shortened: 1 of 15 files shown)


anbenson:
Can you fix the README too?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -98,26 +100,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+      }
+      BufferedReader reader = new BufferedReader(new InputStreamReader(entity.getContent()));
+      StringBuilder responseBody = new StringBuilder();
+      String line;
+      while ((line = reader.readLine()) != null) {
+        responseBody.append(line);
+      }
+      JSONObject json = new JSONObject(responseBody.toString());
+      if (!json.isNull(CSE_ITEMS) && !json.getJSONArray(CSE_ITEMS).isNull(0)) {
+        JSONObject obj = json.getJSONArray(CSE_ITEMS).getJSONObject(0);
```
irisliu77, line 124: we might get 500 error here and it cannot be handled by throwing exception.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -98,26 +100,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+        if (!obj.isNull(CSE_LINK)) {
+          return obj.getString(CSE_LINK);
+        }
+      }
+      return null;
+    } catch (IOException e) {
```
irisliu77, line 130: a long try-catch block is not desired according to @anbenson

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
kaylaleung: Uhhh then how else should we handle it? I tried to do checks that each of the fields exist/not null to avoid an error when accessing a JSON field that doesn't exist?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -102,22 +104,49 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+      while ((line = reader.readLine()) != null) {
+        responseBody.append(line);
+      }
+      reader.close();
+    } catch (IOException e) {
+      return null;
+    }
+    JSONObject json = new JSONObject(responseBody.toString());
+
+    if (response.getStatusLine().getStatusCode() != 200) {
```
anbenson, line 140: You've already done this above; no need to do it again.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -102,22 +104,49 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+    String line;
+    try {
+      while ((line = reader.readLine()) != null) {
+        responseBody.append(line);
+      }
+      reader.close();
+    } catch (IOException e) {
+      return null;
+    }
+    JSONObject json = new JSONObject(responseBody.toString());
```
anbenson, line 138: So I think you can shorten this method by about a third by directly calling `JSONObject.parse()` on your InputStreamReader. No need to create a BufferedReader, or to read into a StringBuilder, etc. You can use a Java 7 try-with-resources idiom to simultaneously catch exceptions and close resources.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -102,22 +106,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
-            .setHeader("User-Agent", "Java 11 HttpClient Bot")
-            .build();
-    HttpResponse<String> linkResponse;
+    String url = CSE_URL + "?key=" + API_KEY + "&cx=" + id + "&q=" + query;
+    CloseableHttpClient httpClient = HttpClients.createDefault();
+    CloseableHttpResponse response;
     try {
-      linkResponse = httpClient.send(linkRequest, HttpResponse.BodyHandlers.ofString());
-    } catch (IOException | InterruptedException ex) {
+      response = httpClient.execute(new HttpGet(url));
```
joanchenx, line 113: @anbenson mentioned using try-with-resources idiom to catch exceptions and close resources.

The top answer demonstrates how to do this at the end of their answer: https://stackoverflow.com/questions/21574478/what-is-the-difference-between-closeablehttpclient-and-httpclient-in-apache-http

I think to do this, you'll have to move more of your logic into the try/catch as you'll need to finish consuming the response before closing the resources. Regardless of whether you use the try-with-resources or not, you'll want to close the resources somehow.

I'm less familiar with this so I don't want to lead you astray - @anbenson could you confirm that my comment is correct (or horribly incorrect) with regards to using try-with-resources?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -102,22 +106,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
-            .setHeader("User-Agent", "Java 11 HttpClient Bot")
-            .build();
-    HttpResponse<String> linkResponse;
-    try {
-      linkResponse = httpClient.send(linkRequest, HttpResponse.BodyHandlers.ofString());
-    } catch (IOException | InterruptedException ex) {
+    String url = CSE_URL + "?key=" + API_KEY + "&cx=" + id + "&q=" + query;
+    try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
+      try (CloseableHttpResponse response = httpClient.execute(new HttpGet(url))) {
+
```
anbenson, line 112: remove empty line

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -102,22 +106,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+        }
+        HttpEntity entity = response.getEntity();
+        if (entity == null) {
+          return null;
+        }
+        JsonParser parser = new JsonParser();
+        JsonObject obj;
+        try {
+          obj = parser.parse(new InputStreamReader(entity.getContent())).getAsJsonObject();
+        } catch (JsonIOException
```
anbenson, line 124: parser.parse is deprecated in 2.8.6 (version being used). Use JsonParser.parseReader.

This static method can also throw JsonParseException.

getAsJsonObject() can also throw IllegalStateException.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -102,22 +106,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+            | JsonSyntaxException
+            | UnsupportedOperationException
+            | IOException e) {
+          return null;
+        }
+        if (!obj.isJsonNull()
+            && !obj.get(CSE_ITEMS).getAsJsonArray().isJsonNull()
```
anbenson, line 131: obj.get can return null if CSE_ITEMS isn't a valid key. Can guard with has()

anbenson: Note that you don't need these isJsonNull() checks. get() will definitely return null if obj was really the JSONObject representing null.

```java
if (obj.has(CSE_ITEMS) && obj.getAsJsonArray(CSE_ITEMS).size() > 0) {
```
is what you're looking for I think. Note how you can straightaway getAsJsonArray(CSE_ITEMS) without doing a get() first.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -102,22 +106,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+            && obj.get(CSE_ITEMS).getAsJsonArray().size() > 0) {
+          JsonObject item = obj.get(CSE_ITEMS).getAsJsonArray().get(0).getAsJsonObject();
```
anbenson, line 133: Ditto here. Though it'd be nice if you could have a more descriptive variable name than `item`.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -102,22 +106,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+          if (!item.isJsonNull() && !item.get(CSE_LINK).isJsonNull()) {
+            return item.get(CSE_LINK).getAsString();
+          }
+        }
+      }
+    } catch (IOException e) {
```
anbenson, line 139: This catch corresponds to the outer try which can't throw an exception (it just creates a HttpClient). This should correspond to the try that executes the HTTP GET.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
anbenson: I am no expert with try-with-resources, but my understanding is that it's like Python's `with open("file.txt", "r") as f:` syntax, so I suppose so.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
kaylaleung: It gets the "item" field of the JSON.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
kaylaleung: There's a HTTP ERROR 500 java.lang.NoSuchMethodError: com.google.gson.JsonParser.parseReader(Lcom/google/gson/stream/JsonReader;)Lcom/google/gson/JsonElement; for using JsonParser.parseReader in 2.8.6

kaylaleung: Switching to 2.8.5 has no issues with being deprecated with the original code. JsonParser.parseReader doesn't exist in 2.8.5 and seems like 2.8.6 can't detect it at run time?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
kaylaleung: The outer try can throw an exception due to the try-with structure apparently. Because `Unhandled exception type IOException thrown by automatic close() invocation on httpClientJava(16778098)`

But yes, will add additional cath for the nested try.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
joanchenx: I think it's working for me? If I switch to 2.8.6 and then modify the line to be:
`obj = JsonParser.parseReader(new InputStreamReader(entity.getContent())).getAsJsonObject();`, I don't see a 500 error. I might be missing something though, is that what you tried?

kaylaleung: Yeah, I used the exact same line, and I get the 500 error :/

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
anbenson: Yeah :/ I certainly don't like it, but I don't see a good way around it this time short of doing manual close() which seems worse.

PR #19, opened 2020-07-11
Integrate w3 and stackoverflow client calls into servlet

kaylaleung:
Call W3SchoolClient and StackOverflowClient from SearchServlet to retrieve cards

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,6 +33,11 @@
       <artifactId>json</artifactId>
       <version>20200518</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>4.5.10</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/pom.xml
+++ b/pom.xml
@@ -3,8 +3,7 @@
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
-  <!-- TODO: set project name. -->
-  <groupId>com.google.step.YOUR_PROJECT_NAME_HERE</groupId>
+  <groupId>com.google.step.snippet</groupId>
   <artifactId>webapp</artifactId>
   <version>1</version>
   <packaging>war</packaging>
@@ -45,8 +44,7 @@
         <artifactId>appengine-maven-plugin</artifactId>
         <version>2.2.0</version>
         <configuration>
-          <!-- TODO: set project ID. -->
-          <deploy.projectId>YOUR_PROJECT_ID_HERE</deploy.projectId>
+          <deploy.projectId>step1-2020</deploy.projectId>
           <deploy.version>1</deploy.version>
         </configuration>
       </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -38,6 +38,11 @@
       <artifactId>httpclient</artifactId>
       <version>4.5.10</version>
     </dependency>
+    <dependency>
+      <groupId>org.jsoup</groupId>
+      <artifactId>jsoup</artifactId>
+      <version>1.13.1</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
+++ b/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
@@ -14,13 +14,11 @@
 
 package com.google.step.snippet.servlets;
 
+import java.io.BufferedReader;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.io.UnsupportedEncodingException;
-import java.net.URI;
 import java.net.URLEncoder;
-import java.net.http.HttpClient;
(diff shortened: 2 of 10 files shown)


anbenson:
This PR is into the `servlet` branch. What is that?

kaylaleung:
It's the newly fixed SearchServlet that's deployable

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -87,7 +96,11 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     String stackLink = getLink(STACK_CSE_ID, query);
     if (stackLink != null) {
       allLinks.add(stackLink);
-      // TODO: Call stackoverflow API to return JSON card content
+      StackOverflowClient soClient = new StackOverflowClient();
+      Card soCard = w3Client.search(soLink);
```
irisliu77, line 100: should be soClient here

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -100,8 +113,11 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
       // TODO: Call scraping function to return JSON card content
     }
 
-    response.setContentType("text/html;");
+    response.setContentType("application/json;");
     response.getWriter().println(allLinks);
+
```
irisliu77, line 118: this line is no longer needed

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -100,8 +113,11 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+    String json = convertToJson(allCards);
+    response.getWriter().println(json);
   }
 
   private String getLink(String id, String query) {
```
irisliu77, line 123: TODO

kaylaleung:
@AaronLopes can you review and let me know which parts might be taken out given no need to convert to JSON now?

pom.xml
```diff
@@ -43,6 +43,11 @@
       <artifactId>jsoup</artifactId>
       <version>1.13.1</version>
     </dependency>
+    <dependency>
```
AaronLopes, line 46: remove this

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -14,6 +14,10 @@
 
 package com.google.step.snippet.servlets;
 
+import com.google.gson.Gson;
```
AaronLopes, line 17: remove

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -144,4 +153,9 @@ private String getLink(String id, String query) {
     }
     return null;
   }
+
+  private String convertToJson(List<Card> cards) {
```
AaronLopes, line 157: remove

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -96,7 +106,6 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
      */
     String geeksLink = getLink(GEEKS_CSE_ID, query);
     if (geeksLink != null) {
```
AaronLopes, line 108: remove conditional, can added in a separate PR w/ the scraper.

kaylaleung: I think we're just keeping this null check consistent across all three sites to ensure that if the link from CSE is null, there's no need to even try creating a card by calling the respective scraping functions. At least afaik this is how Iris and I set up our clients.

AaronLopes: Sure, I just meant that because there is no scraper class yet, there's no reason to do the check rn.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -65,7 +68,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     }
 
     String query = encodeValue(param);
-    List<String> allLinks = new ArrayList<>();
+    List<Card> allCards = new ArrayList<>();
 
     // TODO: after implementing scraping, consider changing getLink calls to a
     // for-loop
```
joanchenx, line 74: Food for thought for when you consolidate the calls into a for-loop, I think it would make sense to create an interface for your Clients, something like:
```
interface Client {
  public String getCseId();
  public Card search();
}
```
And then have each of your W3School, StackOverflow, and GeeksForGeeks clients implement your `Client` interface. That way, you have a standard API for each of your clients and you can just iterate through a list of clients:
```
List<Client> clients = new ArrayList<Client>();
clients.add(new W3SchoolClient());
...

for (Client client : clients)
{
  Card card = client.search(link);
  ...
}
```
Obviously, don't need to do this in this PR but something to keep in mind. I'll add a Github task for this.

kaylaleung:
#27 is already integrating the clients together in servlet and merging into master

PR #25, opened 2020-07-11
update card class to remove setter

irisliu77:
setters and getters are no longer needed in stackoverflow client as well as default constructor


kaylaleung:
👍

irisliu77:
the setters are again added since @AaronLopes said they will be used in search.jsp

joanchenx:
> the setters are again added since @AaronLopes said they will be used in search.jsp

Nit: I think you mean the getters are added again :)

PR #27, opened 2020-07-13
Fix html escaping on card display

AaronLopes:
address issues #28 and #44


src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -12,7 +14,22 @@
...
+            <div class="card-content">
+              <a href="<%=cards.get(0).getLink()%>" title="card-source" target="_blank">
+                <span class="card-title"><%=cards.get(0).getTitle()%></span>
+              </a>
+              <p><%=cards.get(0).getCode()%></p>
+              <span class="card-title">Description</span>
+              <p><%=cards.get(0).getDescription()%></p>
+            </div>
+          </div>
+      </div>
```
irisliu77, line 32: does these line only used to display one card? if so, TODO for adding another two cards.

AaronLopes: yeah, I wanted to check if ya'll are ok with how it looks so far as the other two will follow a similar format.

irisliu77:
lgtm!

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -74,7 +77,10 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
      */
     String w3Link = getLink(W3_CSE_ID, query);
     if (w3Link != null) {
-      allLinks.add(w3Link);
+      W3SchoolClient client = new W3SchoolClient();
```
anbenson, line 80: So I notice that these clients have zero internal state, which makes sense. This makes search() feel like something that should be a static method on the W3SchoolClient (and friends) class.

Alternatively I also like Joan's proposal of making an interface for these clients. My only addition to it is to not instantiate a client on every request - just make a `List<Client>` in the constructor.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -84,20 +90,19 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
      * Google CSE
      */
     String geeksLink = getLink(GEEKS_CSE_ID, query);
-    if (geeksLink != null) {
-      allLinks.add(geeksLink);
-      // TODO: Call scraping function to return JSON card content
-    }
-
+    if (geeksLink != null) {}
+    request.setAttribute("card_list", allCards);
```
anbenson, line 105: `card_list` is supposed to be like a Java variable, so please use camelCase.

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -12,7 +14,25 @@
...
     <div class="srp-container">
-      <div class="gcse-searchresults-only"></div>
+      <div class="srp-child">
+        <div class="gcse-searchresults-only"></div>
+      </div>
+      <div class="srp-child">
+        <%ArrayList<Card> cards = (ArrayList<Card>)request.getAttribute("card_list");%>
+          <div class="card">
+            <div class="card-content">
+              <a href="<%=cards.get(0).getLink()%>" title="card-source" target="_blank">
```
anbenson, line 24: Most people expect the page to navigate to the search result on click, not open a new tab.

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -12,7 +14,25 @@
     <t:base/>
     <div class="gcse-searchbox-only" data-resultsUrl="/search"></div>
     <div class="srp-container">
-      <div class="gcse-searchresults-only"></div>
+      <div class="srp-child">
+        <div class="gcse-searchresults-only"></div>
+      </div>
+      <div class="srp-child">
+        <%ArrayList<Card> cards = (ArrayList<Card>)request.getAttribute("card_list");%>
```
anbenson, line 21: It's not hard to make this a for loop. I did it in the example code, see https://github.com/googleinterns/step1-2020/commit/8a66c41c2ebe9e3ae0000b6b800a40098b801f90#diff-de7f7adba792bdbbac028ec11fe4dc91

It also avoids having to import and declare an `ArrayList`. (I should point out that you don't need an ArrayList here, just a List, and that's why you're being forced to cast - because it's only a List.)

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
irisliu77: done by adding interface and list

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -12,7 +13,23 @@
...
+            <div class="card-content">
+              <a href="${knowledgeCard.getLink()}" title="card-source">
+                <span class="card-title">${knowledgeCard.getTitle()}</span>
+              </a>
+              <p>${knowledgeCard.getCode()}</p>
+              <span class="card-title">Description</span>
+              <p>${knowledgeCard.getDescription()}</p>
+            </div>
+          </div>
+        </c:forEach>
```
irisliu77, line 31: this code doesn't seem to work when i combine all three client on my local sever. Each client works well with your precious code of getting single card.

AaronLopes: reference the comment left in #33

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -12,7 +13,23 @@
...
+      </div>
+      <div class="srp-child">
+        <c:forEach items='${requestScope["cardList"]}' var="knowledgeCard">
+          <div class="card">
+            <div class="card-content">
+              <a href="${knowledgeCard.getLink()}" title="card-source">
+                <span class="card-title">${knowledgeCard.getTitle()}</span>
+              </a>
+              <p>${knowledgeCard.getCode()}</p>
+              <span class="card-title">Description</span>
```
kaylaleung, line 27: style nit: When you hover over "Description", a pointer appears instead of a cursor since this shouldn't be clickable. 

I think this is because you're using "card-title" class here which is the same class as you use for the actual hyperlinked knowledge card title. I think this description span should have a different unique class.

src/main/webapp/style.css
```diff
@@ -1,7 +1,41 @@
...
   text-align: center;
 }
 
 .srp-container {
   display: flex;
 }
+
+.srp-child {
+  flex: 0 0 50%;
+}
```
kaylaleung, line 41: style nit: I think it would look better if it was more 2/3 search results, 1/3 card filling up the page rather than 50/50 right now

style nit: Add margin to .card so that the cards don't cling to the sides of the page.

src/main/webapp/style.css
```diff
@@ -1,7 +1,41 @@
...
+
+.card {
+  box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 3px 1px -2px rgba(0, 0, 0, 0.12), 0 1px 5px 0 rgba(0, 0, 0, 0.2);
+  position: relative;
+  transition: box-shadow .25s;
+  border-radius: 2px;
+}
+
+.card .card-title {
+  font-size: 18px;
```
kaylaleung, line 13: style nit: make the card title font bigger. maybe make the description paragraph font smaller. I don't think the description should overshadow the code snippet section itself

src/main/webapp/style.css
```diff
@@ -1,7 +1,54 @@
...
 }
 
+.srp {
+  width: 80%;
+}
+
 .srp-container {
   display: flex;
-}
+}
```
irisliu77, line 54: missing endline

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -1,4 +1,6 @@
 <%@page contentType="text/html" pageEncoding="UTF-8"%>
+<%@page import="com.google.step.snippet.data.Card"%> 
+<%@page import="org.apache.commons.lang3.StringEscapeUtils"%>
```
irisliu77, line 3: this package cannot be resolved on my dev server, but it works well without importing this package

irisliu77:
looks like the width of the knowledge card is based on the length of its content, which leads to really wide knowledge card when the response is long. I think adding a fix length to either the search result or the knowledge card would help prevent this problem.

src/main/webapp/WEB-INF/templates/search.jsp
AaronLopes: Did you add the dependency to pom.xml?

anbenson:
+1 to Iris' comment. I'm getting a tiny SRP and huge knowledge card. I think it may be confusing why what's happening is happening, so let me try to explain.

When laying out flex items, Flexbox first tries to lay out items according to flex-basis. You didn't set it (which is fairly normal), so for width, the actual width value is looked at. The SRP container is `width: 80%`, and the knowledge card container is `auto` (the default, so it didn't need to be set), which means it gets to be as long as it wants to.

During this initial pass, Flexbox calculated that the SRP container gets 80% width, so 860 pixels on my computer, and the knowledge card container got as much width as it wanted, and since this particular knowledge card is so big, the container wanted 8158.5 pixels (thanks, Firefox dev tools, Chrome can't tell you this). Yeah, that's not a typo.

After the initial pass, Flexbox tries to figure out whether items have to grow or shrink to fit into the flex container, which is in actuality only 1075 pixels wide (for my computer with dev tools open. note that 80% of this is 860px, as above). Since this is smaller than 860px + 8158.5px, things have to shrink. This is controlled by the flex-shrink property on the flex items, which are not set, so it defaults to 1, which means both items equally lose space. But it's not exactly a fair to do this, since the second flex-item can afford to lose so much more width than the first, which is exactly what happens - the first flex-item loses as much width as it can, down to the minimum size, and the rest is taken from the second flex-item until it fits into the container.

I thought it might be good to understand what's going on because I find no one really explains what's going on under the hood for Flexbox, only what the properties mean at a high-level. But the root issue is that you can't let the second flex-item (the knowledge card container) have infinite width in a flexbox - it's going to overpower the other flex-item when it gets to shrinking.

The solution is definitely to set some widths/max-widths/min-widths on these items. Also you might want to consider what you're trying to do with flexbox - do you really need it? How do you want things to be laid out?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -74,8 +77,9 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
      */
     String w3Link = getLink(W3_CSE_ID, query);
     if (w3Link != null) {
-      allLinks.add(w3Link);
-      // TODO: Call scraping function to return JSON card content
+      W3SchoolClient client = new W3SchoolClient();
+      Card w3Card = client.search(w3Link);
+      allCards.add(w3Card);
```
kaylaleung, line 82: Need to add null check here

`if (w3Card != null) {
allCards.add(w3Card);
}`

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -84,20 +88,18 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
      */
     String stackLink = getLink(STACK_CSE_ID, query);
     if (stackLink != null) {
-      allLinks.add(stackLink);
-      // TODO: Call stackoverflow API to return JSON card content
+      StackOverflowClient stackClient = new StackOverflowClient();
+      Card stackCard = stackClient.search(stackLink);
+      allCards.add(stackCard);
```
kaylaleung, line 93: And here `if (stackCard != null) {
allCards.add(stackCard);
}`

src/main/webapp/WEB-INF/templates/search.jsp
kaylaleung: import seems fine on my end

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -12,7 +14,23 @@
...
+        <div class="gcse-searchresults-only"></div>
+      </div>
+      <div class="card-container">
+        <c:forEach items='${requestScope["cardList"]}' var="knowledgeCard">
+          <div class="card">
+            <div class="card-content">
+              <a href="${knowledgeCard.getLink()}" title="card-source">
+                <span class="card-title">${knowledgeCard.getTitle()}</span>
+              </a>
+              <p>${StringEscapeUtils.escapeHtml4(knowledgeCard.getCode())}</p>
```
kaylaleung, line 27: style: change the paragraph to `<code></code>` instead.

src/main/webapp/style.css
```diff
@@ -1,7 +1,57 @@
...
+  display: flex;
+}
+
+.card .card-content {
+  padding: 24px;
+  border-radius: 0 0 2px 2px;
+  width: 400px;
+}
+
+.card .card-content p {
```
kaylaleung, line 41: style: change this to `.card .card-content p, code ` if you change above to `<code>`

src/main/webapp/style.css
```diff
@@ -38,7 +38,7 @@ body, html {
   width: 400px;
 }
 
-.card .card-content p {
+.card .card-content p, code {
   display: flex;
```
irisliu77: the display here seems weird for displaying stackoverflow code snippet

src/main/webapp/style.css
```diff
@@ -1,7 +1,59 @@
...
+.card .card-title {
+  font-size: 20px;
+  font-weight: 300;
+  cursor: pointer;
+  line-height: 32px;
+  display: flex;
+}
+
+.card .card-content {
+  padding: 24px;
```
anbenson, line 36: When looking and .card and .card-content overall, I'm confused by the margins and padding. There's a lot of empty space on the right, created by making .card-content 400px but card/card-container 500px. There's both padding on .card-content and margin-top on .card. I guess some of this will be made obsolete by combining .card and .card-content.

src/main/webapp/style.css
```diff
@@ -1,7 +1,59 @@
...
+
+.card-container {
+  margin-right: 30px;
+  margin-left: 30px;
+  flex-shrink: 0;
+  flex-basis: 500px;
+}
+
+.card .card-description {
+  font-size: 16px;
```
anbenson, line 22: Combine font-size and font-weight into a single font property.

src/main/webapp/style.css
```diff
@@ -1,7 +1,59 @@
...
+  font-weight: bold;
+  line-height: 32px;
```
anbenson, line 24: Don't use line-height to create margins/padding. Instead, use margin/padding.

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -12,7 +13,23 @@
...
-      <div class="gcse-searchresults-only"></div>
+      <div class="srp">
+        <div class="gcse-searchresults-only"></div>
+      </div>
+      <div class="card-container">
+        <c:forEach items='${requestScope["cardList"]}' var="knowledgeCard">
+          <div class="card">
+            <div class="card-content">
+              <a href="${knowledgeCard.getLink()}" title="card-source">
+                <span class="card-title">${knowledgeCard.getTitle()}</span>
```
anbenson, line 24: Just put the class on the link, no need for the span.

src/main/webapp/style.css
```diff
@@ -1,7 +1,59 @@
...
+.card {
+  box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 3px 1px -2px rgba(0, 0, 0, 0.12), 0 1px 5px 0 rgba(0, 0, 0, 0.2);
+  position: relative;
+  transition: box-shadow .25s;
+  border-radius: 2px;
+  margin-top: 30px;
+  width: 500px;
+}
+
+.card-container {
```
anbenson, line 14: Can you put your CSS in a hierarchical order? I'm mixed up by things like this outer rule being in the middle of the card rules.

src/main/webapp/style.css
```diff
@@ -1,7 +1,59 @@
...
+.card .card-description {
+  font-size: 16px;
+  font-weight: bold;
+  line-height: 32px;
+}
+
+.card .card-title {
+  font-size: 20px;
+  font-weight: 300;
+  cursor: pointer;
```
anbenson, line 30: This does nothing, links are already cursor: pointer.

src/main/webapp/style.css
```diff
@@ -1,7 +1,59 @@
...
+}
+
+.card .card-description {
+  font-size: 16px;
+  font-weight: bold;
+  line-height: 32px;
+}
+
+.card .card-title {
+  font-size: 20px;
```
anbenson, line 28: Combine font-size and font-weight.

src/main/webapp/style.css
```diff
@@ -1,7 +1,59 @@
...
+  font-weight: 300;
+  cursor: pointer;
+  line-height: 32px;
+  display: flex;
```
anbenson, line 32: I'm pretty sure you didn't mean this to be a flexbox, given it has 0 children.

src/main/webapp/style.css
```diff
@@ -1,7 +1,59 @@
...
+}
+
+.card .card-content {
+  padding: 24px;
+  border-radius: 0 0 2px 2px;
+  width: 400px;
+  display: flex;
```
anbenson, line 39: It's overkill to make this a flexbox - you just want blocks of text appearing one below the other, which is the default behavior of block elements like <p>. (Incidentally, you overrode the display: block behavior of <p> to inline...)

But making it a flexbox is causing a few weird bugs. For example, if you hover your mouse to the right of the link, you'll notice that the empty space is clickable. This is because the flexbox is making the link extend past the text.

src/main/webapp/style.css
anbenson: (This also means I'd expect .srp and so forth to appear at the start.)

anbenson:
Also, there's a few changes to files that are not related to the PR ("add card display to srp"). Aside from those changes to SearchServlet.java needed to actually display cards, can you move them out of the PR? It'd be better to discuss them separately anyway so we can land this one.

src/main/java/com/google/step/snippet/external/W3SchoolClient.java
```diff
@@ -33,7 +34,8 @@ public Card search(String w3Link) {
     }
     String title = titles.first().text();
     String description = descriptions.first().text();
-    String code = snippets.first().getElementsByClass(CODE_CLASS).text();
+    String code =
```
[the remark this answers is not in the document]
kaylaleung: SO retrieved code snippets already has `<code></code>` tags wrapped around it and escaping the HTML in those cases was making the code tags appear. But yes, likely many other escaping bugs/edge cases to be addressed later.

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -110,6 +110,9 @@ private String getDescription(String body) {
         break;
       }
     }
+    Jsoup.parse(description).text();
```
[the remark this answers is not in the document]
irisliu77: this convert the html string to raw text, which helps better display of description

src/main/webapp/style.css
```diff
@@ -1,7 +1,49 @@
+body, html {
```
joanchenx, line 1: I think you can delete this file now, since you moved the shared CSS to css/style.css instead?

kaylaleung:
Not sure what's happening with this PR / when or what I should review or if it's just waiting for #43 ?

AaronLopes:
> Not sure what's happening with this PR / when or what I should review or if it's just waiting for #43 ?

This PR is waiting on #43, and will also address the misc issues regarding display (eg. escaping html) that do not involve fixing front-end code.

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -1,4 +1,5 @@
 <%@page contentType="text/html" pageEncoding="UTF-8"%>
+<%@page import="com.google.step.snippet.data.Card"%>
```
joanchenx, line 2: Remove this?

joanchenx:
also, maybe edit the PRs title and description to better describe what the PR is changing now?

kaylaleung:
> also, maybe edit the PRs title and description to better describe what the PR is changing now?

+1 
This PR intended to specifically fixing #28 and #44 right? There are multiple ongoing issues with fixing card displaying rn

anbenson:
AFAIU, this isn't quite ready for review, right? The approvals are for a much earlier version of the PR.

kaylaleung:
Wait, is this the most updated fix, or should I wait for forthcoming changes like the ones we discussed about filtering for keywords or URL paths before reviewing this?

AaronLopes:
> ait, is this the most updated fix, or should I wait for forthcoming changes like the ones we discussed about filtering for keywords or URL paths before reviewing this

the latter, the html escaping issues will be addressed in this PR

anbenson:
Yeah same question, this only escapes two clients' `code` fields. Will HTML be stripped from the other fields? How will this interact with the planned (maybe?) changes to how the SO parser will deal with code/description?

AaronLopes:
> Yeah same question, this only escapes two clients' `code` fields. Will HTML be stripped from the other fields? How will this interact with the planned (maybe?) changes to how the SO parser will deal with code/description?

This PR currently contains the old solution to escaping the HTML, it's going to be updated with new solution to html escaping as well as the new design (JSP rendering) required to support the changes being made to the SO client

anbenson:
Sounds like the current changes in this PR are going to be obsolete. In that case, can you make a new PR with those changes?

(AFAIK, the only thing left in this PR will then be the "..." being appended to descriptions?)

irisliu77:
> Sounds like the current changes in this PR are going to be obsolete. In that case, can you make a new PR with those changes?
> 
> (AFAIK, the only thing left in this PR will then be the "..." being appended to descriptions?)

appending ',,,' for SO is also done in other pr.

irisliu77:
> Yeah same question, this only escapes two clients' `code` fields. Will HTML be stripped from the other fields? How will this interact with the planned (maybe?) changes to how the SO parser will deal with code/description?

SO description now has `<code>` for code snippet, so it will be fine if we add style to code tag.

PR #29, opened 2020-07-14
Add w3client unit tests

kaylaleung:
@anbenson @joanchenx I wanted to check if these were the kind of unit testing you had in mind with regards to the scraping clients?

### DIFF
--- a/src/main/java/com/google/step/snippet/data/Card.java
+++ b/src/main/java/com/google/step/snippet/data/Card.java
@@ -46,4 +46,28 @@ public final class Card {
   public String getDescription() {
     return this.description;
   }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (obj == null) {
+      return false;
+    }
+    if (!Card.class.isAssignableFrom(obj.getClass())) {
+      return false;
+    }
+    final Card other = (Card) obj;
+    if (!this.title.equals(other.title)) {
+      return false;
+    }
+    if (!this.code.equals(other.code)) {
+      return false;
+    }
+    if (!this.link.equals(other.link)) {
+      return false;
+    }
+    if (!this.description.equals(other.description)) {
+      return false;
+    }
+    return true;
+  }
 }
--- a/src/main/java/com/google/step/snippet/data/Card.java
+++ b/src/main/java/com/google/step/snippet/data/Card.java
@@ -31,6 +31,9 @@ public final class Card {
 
   @Override
   public boolean equals(Object obj) {
+    if (this == obj) {
+      return true;
+    }
     if (obj == null) {
       return false;
     }
@@ -38,18 +41,10 @@ public final class Card {
       return false;
     }
     final Card other = (Card) obj;
-    if (!this.title.equals(other.title)) {
-      return false;
-    }
-    if (!this.code.equals(other.code)) {
-      return false;
-    }
-    if (!this.link.equals(other.link)) {
-      return false;
-    }
-    if (!this.description.equals(other.description)) {
-      return false;
-    }
-    return true;
+
+    return this.title.equals(other.title)
+        && this.code.equals(other.code)
+        && this.link.equals(other.link)
+        && this.description.equals(other.description);
   }
 }
--- a/src/main/java/com/google/step/snippet/data/Card.java
+++ b/src/main/java/com/google/step/snippet/data/Card.java
@@ -40,8 +40,8 @@ public final class Card {
     if (!Card.class.isAssignableFrom(obj.getClass())) {
       return false;
     }
-    final Card other = (Card) obj;
 
+    final Card other = (Card) obj;
     return this.title.equals(other.title)
         && this.code.equals(other.code)
         && this.link.equals(other.link)
--- a/src/test/java/com/google/step/snippet/W3SchoolClientTest.java
+++ b/src/test/java/com/google/step/snippet/W3SchoolClientTest.java
@@ -1,8 +1,10 @@
 package com.google.step.snippet;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
 import com.google.step.snippet.data.Card;
 import com.google.step.snippet.external.W3SchoolClient;
-import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
@@ -10,7 +12,7 @@ import org.junit.runners.JUnit4;
 @RunWith(JUnit4.class)
 public final class W3SchoolClientTest {
 
-  W3SchoolClient client = new W3SchoolClient();
+  private final W3SchoolClient client = new W3SchoolClient();
 
   @Test
   public void htmlImage() {
@@ -21,7 +23,7 @@ public final class W3SchoolClientTest {
             "<img src=\"img_girl.jpg\" alt=\"Girl in a jacket\" width=\"500\" height=\"600\">",
             "https://www.w3schools.com/tags/tag_img.asp",
             "How to insert an image:");
-    Assert.assertTrue(actual.equals(expected));
+    assertEquals(actual, expected);
   }
 
   @Test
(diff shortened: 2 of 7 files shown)


anbenson:
Yeah, this sort of thing. JUnit is the right tool here. However, the code under test is in the [implicit] constructor, not in a test itself - generally you want it in the test. For example, if there's an exception raised, you're not going to see a test fail, but something more mysterious.

The general format here I'd expect is to create a client in the constructor or setup, and for each test, feed in some kind of W3Schools URL (or non-W3Schools-URL) and compare the output of the method to an expected Card. You can either override equals or make a custom assert to check each field of a Card.

Other notes:
- Why JUnit 4? Why not JUnit 5?
- Can you add how to run the test command to the README?

joanchenx:
To add on a bit, also add in failure cases too, so an invalid W3Schools URL or a W3Schools URL with missing fields like this URL: https://www.w3schools.com/html/html_quiz.asp which doesn't have a w3-code class.

kaylaleung:
> Yeah, this sort of thing. JUnit is the right tool here. However, the code under test is in the [implicit] constructor, not in a test itself - generally you want it in the test. For example, if there's an exception raised, you're not going to see a test fail, but something more mysterious.
> 
> The general format here I'd expect is to create a client in the constructor or setup, and for each test, feed in some kind of W3Schools URL (or non-W3Schools-URL) and compare the output of the method to an expected Card. You can either override equals or make a custom assert to check each field of a Card.
> 
> Other notes:
> 
> * Why JUnit 4? Why not JUnit 5?
> * Can you add how to run the test command to the README?

JUnit 4 was used in the walkthrough testing guide and JUnit 5 has some overhead with more than one required dependancy.

src/main/java/com/google/step/snippet/data/Card.java
```diff
@@ -28,4 +28,28 @@ public String getLink() {
   public String getDescription() {
     return description;
   }
+
+  @Override
+  public boolean equals(Object obj) {
```
joanchenx, line 33: Quick note, at the beginning of your equals method, add a self-check so if the objects are the same, you can return true:
```
if (this == obj) {
  return true;
}
```

src/main/java/com/google/step/snippet/data/Card.java
```diff
@@ -28,4 +28,28 @@ public String getLink() {
...
+    if (!this.code.equals(other.code)) {
+      return false;
+    }
+    if (!this.link.equals(other.link)) {
+      return false;
+    }
+    if (!this.description.equals(other.description)) {
+      return false;
+    }
+    return true;
```
joanchenx, line 53: For these checks, I think you can collapse it into a single return statement. If one check fails, the other checks won't happen anyways so the single return is equivalent to the list of if checks:
```
return this.title.equals(other.title)
     && this.code.equals(other.code)
     && this.link.equals(other.link)
     && this.description.equals(other.description);
```

src/main/java/com/google/step/snippet/data/Card.java
```diff
@@ -28,4 +28,23 @@ public String getLink() {
...
+      return true;
+    }
+    if (obj == null) {
+      return false;
+    }
+    if (!Card.class.isAssignableFrom(obj.getClass())) {
+      return false;
+    }
+    final Card other = (Card) obj;
+
```
joanchenx, line 44: Super nit: I think it makes more sense to have a new line before the `final Card other` line and no new line between `final Card other` and the `return` statement.

src/test/java/com/google/step/snippet/W3SchoolClientTest.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.step.snippet.external.W3SchoolClient;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class W3SchoolClientTest {
+
+  W3SchoolClient client = new W3SchoolClient();
```
anbenson, line 13: `private final`

src/test/java/com/google/step/snippet/W3SchoolClientTest.java
```diff
@@ -0,0 +1,60 @@
+package com.google.step.snippet;
+
+import com.google.step.snippet.data.Card;
+import com.google.step.snippet.external.W3SchoolClient;
+import org.junit.Assert;
```
anbenson, line 5: By convention, we statically import junit assert methods

e.g. `import static org.junit.Assert.assertEquals;`

src/test/java/com/google/step/snippet/W3SchoolClientTest.java
```diff
@@ -0,0 +1,60 @@
...
+  @Test
+  public void htmlImage() {
+    Card actual = client.search("https://www.w3schools.com/tags/tag_img.asp");
+    Card expected =
+        new Card(
+            "HTML <img> Tag",
+            "<img src=\"img_girl.jpg\" alt=\"Girl in a jacket\" width=\"500\" height=\"600\">",
+            "https://www.w3schools.com/tags/tag_img.asp",
+            "How to insert an image:");
+    Assert.assertTrue(actual.equals(expected));
```
anbenson, line 24: Use assertEquals. Always go for the most semantic assert so that failing test errors are more descriptive e.g. it'll say "foo is not equal to bar as expected" instead of "this boolean is not true as expected"

src/test/java/com/google/step/snippet/W3SchoolClientTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class W3SchoolClientTest {
+
+  W3SchoolClient client = new W3SchoolClient();
+
+  @Test
+  public void htmlImage() {
```
anbenson, line 16: Can you make the test method more descriptive as to what you're testing / write some JavaDoc? I am not sure the rationale behind these particular links.

src/test/java/com/google/step/snippet/W3SchoolClientTest.java
```diff
@@ -0,0 +1,60 @@
...
+            "https://www.w3schools.com/js/js_json_intro.asp",
+            "JSON: JavaScript Object Notation.");
+    Assert.assertTrue(actual.equals(expected));
+  }
+
+  @Test
+  public void incompleteCard() {
+    Card actual = client.search("https://www.w3schools.com/html/html_quiz.asp");
+    Card expected = null;
+    Assert.assertEquals(actual, expected);
```
anbenson, line 44: I think there's an assertNull

src/test/java/com/google/step/snippet/W3SchoolClientTest.java
kaylaleung: Tried to make the test methods more descriptive. Not sure if additional JavaDoc is needed now?

src/test/java/com/google/step/snippet/W3SchoolsClientTest.java
```diff
@@ -0,0 +1,59 @@
...
+  @Test
+  public void htmlCodeCard() {
+    Card actual = client.search("https://www.w3schools.com/tags/tag_img.asp");
+    Card expected =
+        new Card(
+            "HTML <img> Tag",
+            "<img src=\"img_girl.jpg\" alt=\"Girl in a jacket\" width=\"500\" height=\"600\">",
+            "https://www.w3schools.com/tags/tag_img.asp",
+            "How to insert an image:");
+    assertEquals(actual, expected);
```
anbenson, line 26: http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(java.lang.Object,%20java.lang.Object)

The order is actually `assertEquals(expected, actual)`. This only matters to make the error message read correctly. (and so forth for the others)

PR #36, opened 2020-07-16
Add Login/Logout button to homepage and search page

kaylaleung:
Auth class methods can be called from future servlets.


src/main/java/com/google/step/snippet/data/Auth.java
```diff
@@ -0,0 +1,25 @@
+package com.google.step.snippet.data;
+
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+
+public class Auth {
```
anbenson, line 6: This class feels like a bit of needless indirection.

If you didn't use this Auth class, you would keep a UserService around. (Note that you don't even need to - UserServiceFactory.getUserService() is using the Singleton design pattern, meaning that after the first call to it, subsequent calls to it are essentially free, so you can call it as much as you want without performance loss.)

If you do use this Auth class, you need to keep an instance of it around. But all the methods are just wrappers, so I don't see any benefit.

kaylaleung: I edited the class so that it's no longer just wrappers. This reduces the need to the if checks for login within the servlet itself since the Auth class will take care of it and return the information that the servlet is _actually_ looking for since usually, the servlet will want to *do something* if the user is authenticated, not just checking _if_ the user is authenticated. I think this class will be useful for future features that will need to check if the user is logged in to do something.

src/main/java/com/google/step/snippet/servlets/HomepageServlet.java
```diff
@@ -25,9 +26,19 @@
...
+  private static final String AUTH_URL = "authUrl";
+  private static final String AUTH_LABEL = "authLabel";
+
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response)
       throws IOException, ServletException {
+    String redirectPath = request.getRequestURI();
+    if (request.getQueryString() != null) {
+      redirectPath += "?" + request.getQueryString();
+    }
```
anbenson, line 38: I am confused what this does. What does request.getQueryString() do? What is the intent of this code?

src/main/java/com/google/step/snippet/data/Auth.java
anbenson: I can see the purpose of a shared class to handle retrieving the button url and text. But I am unconvinced there will be many other things to share. For example, only the feedback servlet will care about checking the user is authenticated prior to incrementing a vote tally.

I am also unconvinced of the statement that "the servlet will want to do something if the user is authenticated, not just checking if the user is authenticated". For example, the feedback servlet will want to do nothing if asked to upvote but the user is not authenticated (maybe it'll send back an error message). The account servlet will want to display an error page, but that's the job of the servlet, not a shared class.

Can you provide examples of what you're envisioning?

src/main/webapp/style.css
```diff
@@ -5,3 +5,33 @@ h3 {
 .srp-container {
   display: flex;
 }
+
+#auth {
```
anbenson, line 9: Google CSS style prohibits ID selectors due to precedence ambiguity.

src/main/webapp/style.css
```diff
@@ -5,3 +5,33 @@ h3 {
...
+  width: 47px;
+  height: 15px;
+  color: white;
+  background: #4d90fe;
+  border: 1px #3079ED solid;
+  padding: 5px 10px;
+  border-radius: 2px;
+  font: 400 13.3333px Arial;
```
anbenson, line 17: I am dubious about 13.3333. Was 13 or 14 really not good enough?

src/main/webapp/style.css
```diff
@@ -5,3 +5,33 @@ h3 {
...
+  text-align: center;
+  text-decoration: none;
+  font-size: 9pt;
+  outline: none;
```
anbenson, line 21: Why? This looks like it's disabling the focus ring that keyboard navigation users depend on.

src/main/webapp/style.css
```diff
@@ -5,3 +5,33 @@ h3 {
...
+  height: 15px;
+  color: white;
+  background: #4d90fe;
+  border: 1px #3079ED solid;
+  padding: 5px 10px;
+  border-radius: 2px;
+  font: 400 13.3333px Arial;
+  text-align: center;
+  text-decoration: none;
+  font-size: 9pt;
```
anbenson, line 20: You already set the font size.

src/main/webapp/style.css
```diff
@@ -5,3 +5,33 @@ h3 {
 .srp-container {
   display: flex;
 }
+
+#auth {
+  width: 47px;
```
anbenson, line 10: I didn't check, but this feels like you're hardcoding the height and width to fit the size of the text on your browser on your OS on your device. Why not just let it be the size that it renders at?

src/main/webapp/style.css
```diff
@@ -5,3 +5,33 @@ h3 {
...
+  box-shadow: 0 1px 1px #EAEAEA, inset 0 1px 0 #5A94F1;
+  background: #3F83F1;
+}
+
+#auth:active {
+  box-shadow: inset 0 2px 5px #2370FE;   
+}
+
+.topcorner{
+  position:absolute;
```
anbenson, line 35: I don't think this requires absolute positioning. I also wonder if you might want to the logo and sign-in button on the same level. Maybe make a flexbox container for both?

src/main/java/com/google/step/snippet/servlets/HomepageServlet.java
kaylaleung: request.getQueryString() pulls everything after the ? on the path. So in the case of /search?q=html it will pull q=html which I need to construct the full link to redirect to after. But I probably actually don't need this code specifically on the HomepageServlet which will always be '/'

src/main/java/com/google/step/snippet/data/Auth.java
kaylaleung: Hmm I guess I had been thinking that the class would handle the error message for non authenticated user, but making the servlet class instead of the class do that makes sense. 

I this case, I can just keep the class for the button and text which will be shared across the Homepage, Search page, (and future accounts page)?

src/main/webapp/style.css
```diff
@@ -5,3 +5,32 @@ h3 {
 .srp-container {
   display: flex;
 }
+
+.header-container {
+  display: flex;
+  align-items: center;
+  justify-content: space-between; 
+  margin-left: 50%;
```
kaylaleung, line 13: I couldn't figure out a different way to get the logo naming to be in the center, but this also doesn't seem like the most ideal way...

src/main/webapp/css/style.css
```diff
@@ -5,3 +5,36 @@ body, html {
 h3 {
   text-align: center;
 }
+
+h3 {
+  text-align: center;
+}
```
joanchenx, line 11: I think this is repeated from above? (Maybe caused by merging master)

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -63,6 +67,20 @@ private static String encodeValue(String value) {
...
+      redirectPath += "?" + request.getQueryString();
+    }
+
+    UserService userService = UserServiceFactory.getUserService();
+    if (userService.isUserLoggedIn()) {
+      request.setAttribute(AUTH_URL, userService.createLogoutURL(redirectPath));
+      request.setAttribute(AUTH_LABEL, "Logout");
+    } else {
+      request.setAttribute(AUTH_URL, userService.createLoginURL(redirectPath));
+      request.setAttribute(AUTH_LABEL, "Login");
```
irisliu77, line 81: change login/logout to constant

anbenson: I'm actually okay with it here because it's not like the other cases - this is the only place it should ever be used in this file.

(In the other cases where I've asked strings to be made private constants, it was because it was being used as a constant identifier in multiple places, such as as a query parameter or a database key, etc.)

src/main/webapp/WEB-INF/tags/base.tag
```diff
@@ -1,2 +1,5 @@
 <%@tag description="Common page content" pageEncoding="UTF-8"%>
+<div class="header-container">
 <h1><span class="blue">S</span><span class="red">n</span><span class="yellow">i</span><span class="blue">p</span><span class="green">p</span><span class="red">e</span><span class="yellow">t</span></h1>
+<a class="auth-button" id="auth" href="${authUrl}">${authLabel}</a>
+</div>
```
AaronLopes, line 5: add newline

src/main/webapp/WEB-INF/tags/base.tag
```diff
@@ -1,2 +1,5 @@
 <%@tag description="Common page content" pageEncoding="UTF-8"%>
+<div class="header-container">
 <h1><span class="blue">S</span><span class="red">n</span><span class="yellow">i</span><span class="blue">p</span><span class="green">p</span><span class="red">e</span><span class="yellow">t</span></h1>
```
anbenson, line 3: Please indent your HTML normally. I understand you can't put spaces between the spans, but you can still indent the h1 and a as normal.

src/main/webapp/WEB-INF/tags/base.tag
```diff
@@ -1,2 +1,5 @@
...
+<a class="auth-button" id="auth" href="${authUrl}">${authLabel}</a>
```
anbenson, line 4: I don't see this id being used anywhere.

src/main/webapp/css/style.css
```diff
@@ -22,3 +22,32 @@ h1 {
...
+
+.header-container {
+  display: flex;
+  align-items: center;
+  justify-content: space-between; 
+  margin-left: 48%;
+}
+
+.auth-button {
+  right: 8px;
```
anbenson, line 34: This does nothing since position is not set.

src/main/webapp/css/style.css
```diff
@@ -22,3 +22,32 @@ h1 {
...
+  font: 400 13px Arial;
+  color: white;
+  background: #4d90fe;
+  border: 1px #3079ED solid;
+  padding: 5px 10px;
+  border-radius: 2px;
+  text-align: center;
```
anbenson, line 41: This does nothing since a is an inline element, and text-align works on block elements (and its descendants).

src/main/webapp/css/style.css
```diff
@@ -22,3 +22,32 @@ h1 {
 .green {
   color: #0F9D58;
 }
+
+.header-container {
+  display: flex;
+  align-items: center;
+  justify-content: space-between; 
+  margin-left: 48%;
```
anbenson, line 30: I am very suspicious of this layout :/ but as requested, I'll delay the rest of my judgment to the PR that changes all this CSS.

PR #41, opened 2020-07-17
Add logo style

kaylaleung:
The logo HTML is on one entire line because when I try to shorten the line length, there are spaces between the letters because of the spans... Not sure how to resolve otherwise. :/


src/main/webapp/WEB-INF/tags/base.tag
```diff
@@ -1,2 +1,2 @@
 <%@tag description="Common page content" pageEncoding="UTF-8"%>
-<h3>Snippet</h3>
+<h3><span class="blue">S</span><span class="red">n</span><span class="yellow">i</span><span class="blue">p</span><span class="green">p</span><span class="red">e</span><span class="yellow">t</span></h3>
```
irisliu77, line 2: i think maybe use an img/png as logo instead of hard coding it in html?

kaylaleung:
Re: comment about using image. I'd rather do the logo in HTML than have to deal with responsiveness and resolution issues with an image logo?

irisliu77:
yes i talked to andrew yesterday and he thinks is fine to use either, so i think i delete the comment that i left about using img

src/main/webapp/WEB-INF/tags/base.tag
irisliu77, line 2: should be using `h1` here

src/main/webapp/WEB-INF/tags/base.tag
```diff
@@ -1,2 +1,2 @@
 <%@tag description="Common page content" pageEncoding="UTF-8"%>
-<h3>Snippet</h3>
+<h1><span class="blue">S</span><span class="red">n</span><span class="yellow">i</span><span class="blue">p</span><span class="green">p</span><span class="red">e</span><span class="yellow">t</span></h1>
```
irisliu77, line 2: only thing i'm not sure about is the format here.

kaylaleung: Yeah, there are not intended white spaces between the spans also when I use normal HTML format here :/

PR #42, opened 2020-07-17
Parallelize search servlet

AaronLopes:
create threads to retrieve links and run respective scrapers

### DIFF
--- a/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
+++ b/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
@@ -27,8 +27,8 @@ import java.io.InputStreamReader;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CancellationException;
@@ -37,6 +37,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import javax.servlet.ServletException;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
@@ -62,9 +63,11 @@ public class SearchServlet extends HttpServlet {
   private static final String CARD_LIST_LABEL = "cardList";
   private static final int THREAD_COUNT = 3;
 
-  private final Client w3Client = new W3SchoolsClient(W3_CSE_ID);
-  private final Client stackClient = new StackOverflowClient(STACK_CSE_ID);
-  private final Client geeksClient = new GeeksForGeeksClient(GEEKS_CSE_ID);
+  private final List<Client> clients =
+      Arrays.asList(
+          new W3SchoolsClient(W3_CSE_ID),
+          new StackOverflowClient(STACK_CSE_ID),
+          new GeeksForGeeksClient(GEEKS_CSE_ID));
 
   private static String encodeValue(String value) {
     try {
@@ -84,52 +87,39 @@ public class SearchServlet extends HttpServlet {
       return;
     }
     String query = encodeValue(param);
-    List<Card> allCards = new ArrayList<>();
+    List<Card> allCards;
     ExecutorService executor = Executors.newFixedThreadPool(THREAD_COUNT);
     List<Callable<Card>> cardCallbacks =
-        Arrays.asList(
-            () -> {
-              String w3Link = getLink(w3Client.getCseId(), query);
-              if (w3Link != null) {
-                return w3Client.search(w3Link);
-              }
-              return null;
-            },
-            () -> {
-              String stackLink = getLink(stackClient.getCseId(), query);
-              if (stackLink != null) {
-                return stackClient.search(stackLink);
-              }
-              return null;
-            },
-            () -> {
-              String geeksLink = getLink(geeksClient.getCseId(), query);
-              if (geeksLink != null) {
-                return geeksClient.search(geeksLink);
-              }
-              return null;
-            });
+        clients.stream()
+            .map(
+                client ->
+                    ((Callable<Card>)
+                        () -> {
+                          String link = getLink(client.getCseId(), query);
+                          if (link != null) {
+                            return client.search(link);
+                          }
+                          return null;
+                        }))
+            .collect(Collectors.toList());
     /* thread process completion takes roughly 1.5 to 2.5 seconds, thus a timeout of
      * 3 seconds was chosen.
      */
     try {
-      executor.invokeAll(cardCallbacks, 3L, TimeUnit.SECONDS).stream()
-          .map(
-              future -> {
-                try {
-                  return future.get();
-                } catch (CancellationException | ExecutionException | InterruptedException e) {
-                  return null;
-                }
-              })
-          .forEach(
-              (card) -> {
-                if (card != null) {
-                  allCards.add(card);
-                }
-              });
-    } catch (InterruptedException | NullPointerException | RejectedExecutionException e) {
-      request.getRequestDispatcher("WEB-INF/templates/search.jsp").forward(request, response);
+      allCards =
+          executor.invokeAll(cardCallbacks, 3L, TimeUnit.SECONDS).stream()
+              .map(
+                  future -> {
+                    try {
+                      return future.get();
+                    } catch (CancellationException | ExecutionException | InterruptedException e) {
+                      return null;
+                    }
+                  })
+              .filter(card -> card != null)
+              .collect(Collectors.toList());
+    } catch (InterruptedException | RejectedExecutionException e) {
+      allCards = Collections.emptyList();
     }
 
     request.setAttribute(CARD_LIST_LABEL, allCards);
--- a/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
+++ b/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
@@ -14,6 +14,9 @@
 
 package com.google.step.snippet.servlets;
 
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonParser;
 import com.google.step.snippet.data.Card;
 import com.google.step.snippet.external.GeeksForGeeksClient;
 import com.google.step.snippet.external.StackOverflowClient;
--- a/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
+++ b/src/main/java/com/google/step/snippet/servlets/SearchServlet.java
@@ -77,31 +77,31 @@ public class SearchServlet extends HttpServlet {
 
     ExecutorService executor = Executors.newFixedThreadPool(3);
     List<Callable<Card>> cardCallbacks =
-      Arrays.asList(
-          () -> {
-            String w3Link = getLink(W3_CSE_ID, query);
-            if (w3Link != null) {
-              W3SchoolClient w3client = new W3SchoolClient();
-              return w3client.search(w3Link);
-            }
-            return null;
-          },
-          ()  -> {
-            String stackLink = getLink(STACK_CSE_ID, query);
-            if (stackLink != null) {
-              StackOverflowClient stackClient = new StackOverflowClient();
-              return stackClient.search(stackLink);
-            }
-            return null;
-          },
-          () -> {
-            String geeksLink = getLink(GEEKS_CSE_ID, query);
-            if (geeksLink != null) {
-              GeeksForGeeksClient geeksClient = new GeeksForGeeksClient();
-              return geeksClient.search(geeksLink);
-            }
-            return null;
-          });
+          Arrays.asList(
+              () -> {
+                String w3Link = getLink(W3_CSE_ID, query);
+                if (w3Link != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -66,41 +78,56 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+                return stackClient.search(stackLink);
+              }
+              return null;
+            },
+            () -> {
+              String geeksLink = getLink(GEEKS_CSE_ID, query);
+              if (geeksLink != null) {
+                GeeksForGeeksClient geeksClient = new GeeksForGeeksClient();
+                return geeksClient.search(geeksLink);
+              }
```
kaylaleung, line 107: AFAIK most of this code for calling each of the client's respectively won't be relevant after Iris' Client interface PR since everything will be in a for loop?

AaronLopes: The Client interface could still be used within each Callable, but I don't think a for loop would be needed here

irisliu77:
i think it would be nice to add a callback time to each callable, the estimated time can get from printing the timestamp

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -79,13 +84,50 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+            },
+            () -> {
+              String geeksLink = getLink(geeksClient.getCseId(), query);
+              if (geeksLink != null) {
+                return geeksClient.search(geeksLink);
+              }
+              return null;
+            });
+    try {
+      executor.invokeAll(cardCallbacks, 3L, TimeUnit.SECONDS).stream()
```
irisliu77, line 112: add a comment explaining the call back time/unit

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -79,13 +84,50 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     }
     String query = encodeValue(param);
     List<Card> allCards = new ArrayList<>();
-    for (Client client : clients) {
-      String link = getLink(client.getCseId(), query);
-      Card card = client.search(link);
-      if (card != null) {
-        allCards.add(card);
-      }
+    ExecutorService executor = Executors.newFixedThreadPool(3);
```
kaylaleung, line 87: Set this magic number to be a constant.

Otherwise, lgtm

kaylaleung: Not resolved?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -79,13 +84,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+          .map(
+              future -> {
+                try {
+                  return future.get();
+                } catch (CancellationException | ExecutionException | InterruptedException e) {
+                  return null;
+                }
+              })
+          .forEach(
+              (c) -> {
```
kaylaleung, line 125: Change 'c' to be a more descriptive variable

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -79,13 +84,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+                if (c != null) {
+                  allCards.add(c);
+                }
+              });
+    } catch (InterruptedException | NullPointerException | RejectedExecutionException e) {
+      return;
     }
```
kaylaleung, line 132: I don't think it should be returning here? If an error is caught, SearchServlet should still redirect to the SRP page just with no cards.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -79,13 +84,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+            });
+    /* thread process completion takes roughly 1.5 to 2.5 seconds, thus a timeout of
+     * 3 seconds was chosen.
+     */
+    try {
+      executor.invokeAll(cardCallbacks, 3L, TimeUnit.SECONDS).stream()
+          .map(
+              future -> {
+                try {
+                  return future.get();
```
kaylaleung, line 119: I don't really understand what's being returned here?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -79,13 +84,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+                } catch (CancellationException | ExecutionException | InterruptedException e) {
+                  return null;
+                }
+              })
+          .forEach(
```
kaylaleung, line 124: as far as I understand... map is used to apply a function onto every single element in an list. So I'm not sure why there needs to be a sequential .forEach applied on top of the map?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
kaylaleung: Ohhh wait, future.get() is 'c' here right?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -79,13 +84,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+     * 3 seconds was chosen.
+     */
+    try {
+      executor.invokeAll(cardCallbacks, 3L, TimeUnit.SECONDS).stream()
+          .map(
+              future -> {
+                try {
+                  return future.get();
+                } catch (CancellationException | ExecutionException | InterruptedException e) {
+                  return null;
```
kaylaleung, line 121: Shouldn't be returning null here either?

AaronLopes: The code here is performing an operation (future.get()) on each Future in the list of Future's returned by invokeAll(). I'd say returning null is appropriate here if this operation fails on one of the Future's, because we'd want it to continue through the list.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -97,13 +103,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
     String query = encodeValue(param);
     List<Card> allCards = new ArrayList<>();
-    for (Client client : clients) {
-      String link = getLink(client.getCseId(), query);
-      Card card = client.search(link);
-      if (card != null) {
-        allCards.add(card);
-      }
+    ExecutorService executor = Executors.newFixedThreadPool(THREAD_COUNT);
+    List<Callable<Card>> cardCallbacks =
```
anbenson, line 107: So there's a lot of duplicated code here. My advice (warning: I only somewhat tried compiling this):

```java
List<Callable<Card>> cardCallbacks = clients.stream().map(client -> ((Callable<Card>)() -> {
  if (link != null) {
    return client.search(link);
  }
  return null;
})).collect(Collectors.toList());
```

This maps the list of clients to a list of lambdas that return cards.

Then you can avoid mutating allCards, you can go all-in with functional programming:

```java
List<Card> cards;
try {
  cards = executor.invokeAll(...).stream().map(...).filter(card -> card != null).collect(Collectors.toList());
} catch (InterruptedException | RejectedExecutionException e) {
  cards = Collections.emptyList();
}
```

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -97,13 +103,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+              if (geeksLink != null) {
+                return geeksClient.search(geeksLink);
+              }
+              return null;
+            });
+    /* thread process completion takes roughly 1.5 to 2.5 seconds, thus a timeout of
+     * 3 seconds was chosen.
+     */
+    try {
+      executor.invokeAll(cardCallbacks, 3L, TimeUnit.SECONDS).stream()
```
anbenson, line 134: I would base the timeout on what would be reasonable for a user to wait for results, not how long it's necessary to get results on your machine. (Remember it might take a different amount of time on someone else's machine, or dependent on their internet connection! Although it's probably (?) faster on a server since it has ethernet.)

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -97,13 +103,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+                  return null;
+                }
+              })
+          .forEach(
+              (card) -> {
+                if (card != null) {
+                  allCards.add(card);
+                }
+              });
+    } catch (InterruptedException | NullPointerException | RejectedExecutionException e) {
```
anbenson, line 149: Docs say NPE only happens if you pass in null, which you know you're not doing, so it's fine not to catch.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -97,13 +103,53 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+      request.getRequestDispatcher("WEB-INF/templates/search.jsp").forward(request, response);
```
anbenson, line 150: I wouldn't do this (see suggested code), since I think the JSP template might crash if CARD_LIST_LABEL attribute isn't set.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
AaronLopes: Gotcha, still a little unfamiliar with functional programming, I'll try it out.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
AaronLopes: What would be the best way to go about estimating this? "What's reasonable" seems like a pretty wide range.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -96,14 +105,41 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
     String query = encodeValue(param);
-    List<Card> allCards = new ArrayList<>();
-    for (Client client : clients) {
-      String link = getLink(client.getCseId(), query);
-      Card card = client.search(link);
-      if (card != null) {
-        allCards.add(card);
-      }
+    List<Card> allCards;
+    ExecutorService executor = Executors.newFixedThreadPool(THREAD_COUNT);
```
anbenson, line 109: So I probably should have thought about this earlier, but part of the overhead of threads is creating new threads, and we're doing that on every single request, which is probably part of why you're unlikely to see 3x speedups. My new recommendation (sorry) is to use Executors.newCachedThreadPool(), and make it a private instance variable. This way new requests reuse already existing threads whenever possible, making those requests complete faster, and we don't end up spinning up and tearing 3 threads on every request.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -96,14 +105,41 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
     }
     String query = encodeValue(param);
-    List<Card> allCards = new ArrayList<>();
-    for (Client client : clients) {
-      String link = getLink(client.getCseId(), query);
-      Card card = client.search(link);
-      if (card != null) {
-        allCards.add(card);
-      }
+    List<Card> allCards;
```
anbenson, line 108: Declare this right before you assign to it.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
anbenson: Fair. There's HCI studies that discuss how long a user is willing to wait, but most of the studies boil down to "faster is better so just be as fast as possible". 3 is fine I suppose, as long as it usually enough on your computer.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
anbenson: For reference: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html#newCachedThreadPool()

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -96,14 +105,41 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+                    ((Callable<Card>)
+                        () -> {
+                          String link = getLink(client.getCseId(), query);
+                          if (link != null) {
+                            return client.search(link);
+                          }
+                          return null;
+                        }))
+            .collect(Collectors.toList());
+    /* thread process completion takes roughly 1.5 to 2.5 seconds, thus a timeout of
```
anbenson, line 123: nit: if you decide to keep the comment, can you format the comment like this

```java
// blah blah blah
// blah blah
```

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -57,13 +65,16 @@
   private static final String AUTH_LABEL = "authLabel";
   private static final String CSE_URL = "https://www.googleapis.com/customsearch/v1";
   private static final String CARD_LIST_LABEL = "cardList";
+  private static final int THREAD_COUNT = 3;
```
anbenson, line 68: This constant is unused

PR #43, opened 2020-07-17
Add card display to srp (updated)

AaronLopes:
new PR to consolidate front end changes


irisliu77:
is #27 outdated?

AaronLopes:
> is #27 outdated?

no, there are still features unique to #27 that are needed in the future

src/main/webapp/css/srp-style.css
```diff
@@ -0,0 +1,41 @@
...
+  display: flex;
+}
+
+.srp {
+  flex-grow: 1;
+}
+
+.card-container {
+  margin: 20px 30px 0px 30px;
+  flex: 0 0 400px;
```
anbenson, line 11: I don't think you explicitly need to set flex-basis to 400px since the default value is auto, which will infer the 400px from its contents (e.g. the cards) which are 400px. It's usually best not to duplicate values in case they change. But of course check this first.

src/main/webapp/css/srp-style.css
```diff
@@ -0,0 +1,41 @@
+.srp-container {
```
anbenson, line 1: Can you rename srp-style.css to srp_style.css? Filenames should be lowercase_underscore.

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -1,18 +1,34 @@
 <%@page contentType="text/html" pageEncoding="UTF-8"%>
+<%@page import="com.google.step.snippet.data.Card"%>
```
anbenson, line 2: AFAICT, this isn't used.

PR #52, opened 2020-07-20
Add site source icon 

kaylaleung:
Add image icon associating each card with its original source website


src/main/java/com/google/step/snippet/external/GeeksForGeeksClient.java
```diff
@@ -52,6 +52,6 @@ public Card search(String geeksLink) {
     String title = titles.first().text();
     String description = descriptions.first().text();
     String code = snippets.first().getElementsByClass(CODE_CLASS).text();
-    return new Card(title, code, geeksLink, description);
+    return new Card(title, code, geeksLink, description, "GeeksForGeeks");
```
joanchenx, line 55: Thoughts on making "GeekForGeeks" and the other source names constants?

I know in #36, @anbenson mentioned that "Login" and "Logout" don't need to be constants since it's only used in one place in the file. However, this case feels _slightly_ different since from looking at `new Card(title, code, geeksLink, description, "GeeksForGeeks")` it isn't 100% clear what "GeeksForGeeks" means to Card. I think `new Card(title, code, geeksLink, description, SOURCE_NAME)` or something might make it easier to understand?

Whereas in #36, `request.setAttribute(AUTH_LABEL, "Login")` makes it quite clear what "Login" is being used for.

src/main/java/com/google/step/snippet/data/Card.java
```diff
@@ -29,6 +31,10 @@ public String getDescription() {
     return description;
   }
 
+  public String getSource() {
+    return source;
+  }
+
   @Override
   public boolean equals(Object obj) {
```
anbenson, line 39: Update equals() if you add source

src/main/webapp/css/srp_style.css
```diff
@@ -43,3 +43,11 @@
   margin-top: 0;
   display: block;
 }
+
+.card-heading {
+  display: flex;
```
anbenson, line 48: Flexbox seems overkill. You just have two inline elements next to each other. If you're trying to do vertical alignment, use the `vertical-align` property on the image: https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align

Here, I think you want `vertical-align: bottom;`

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -18,9 +18,12 @@
       <div class="card-container">
         <c:forEach items='${requestScope["cardList"]}' var="knowledgeCard">
           <div class="card">
-            <a href="${knowledgeCard.getLink()}" title="card-source"class="card-title">
-              ${knowledgeCard.getTitle()}
-            </a>
+            <span class="card-heading">
```
anbenson, line 21: We should usually avoid &lt;span&gt; unless there is no better semantic element. Here, &lt;header&gt; seems reasonable.

src/main/java/com/google/step/snippet/external/GeeksForGeeksClient.java
```diff
@@ -11,6 +11,7 @@
   private static final String DESC_TAG = "p";
   private static final String SNIPPET_CLASS = "code-block";
   private static final String CODE_CLASS = "code-container";
+  private static final String SOURCE_NAME = "GeeksForGeeks";
```
anbenson, line 14: I don't think we should implicitly tie these strings with the names of files. If you want this to be the name of a file, be explicit and say "geeksforgeeks.png".

However, I don't think we can add these images into the git repo. We don't own them, and we don't have the license to use them like that. Instead, can we link to the icon itself? For example, GeeksForGeeks' is at https://www.geeksforgeeks.org/favicon.ico

src/main/java/com/google/step/snippet/external/GeeksForGeeksClient.java
```diff
@@ -11,6 +11,8 @@
...
+  private static final String ICON_LINK = "https://geeksforgeeks.org/favicon.ico";
```
anbenson, line 15: Turns out my adblocker, uBlock Origin, blocks favicons used like this. Who knew? Oh well.

Issue #39, opened 2020-07-16
500 Error JsonException from StackOverflow Client


kaylaleung:
When querying for `https://www.w3schools.com/html/html_quotation_elements.asp` in the search box, the redirect link 
`/search?q=https%3A%2F%2Fwww.w3schools.com%2Fhtml%2Fhtml_quotation_elements.asp`throws a HTTP ERROR 500 org.json.JSONException: JSONArray[0] not found from the StackOverflow Client

_At least I'm pretty sure it's StackOverflow Client since when I comment out the StackOverflow .search code this error doesn't appear_

irisliu77:
this because the CSE returns an Stackoverflow question that is not answered yet. need to be resolved in future pr.

irisliu77:
#54

PR #54, opened 2020-07-21
Fix 500 error when stackoverflow post is unanswered

irisliu77:


src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -173,7 +174,11 @@ private String getResponse(String url, String fieldParam) {
       return null;
     }
     JSONObject json = new JSONObject(responseBody.toString());
-    String res = json.getJSONArray(ITEM_PARAMETER).getJSONObject(0).get(fieldParam).toString();
+    JSONArray items = json.getJSONArray(ITEM_PARAMETER);
+    if (items == null || items.length() == 0) {
+      return null;
+    }
+    String res = items.getJSONObject(0).get(fieldParam).toString();
     if (response.getStatusLine().getStatusCode() != 200) {
```
kaylaleung, line 182: Can follow a similar style in SearchServlet #18 :

```
if (json.has(ITEM_PARAMTER) && json.getAsJsonArray(ITEM_PARAMTER).size() > 0) {
    JsonObject item = obj.getAsJsonArray(ITEM_PARAMTER).get(0).getAsJsonObject();
    if (item.has(fieldParam)) {
        String res = item.get(fieldParam).getAsString();
    }
}
```

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -173,7 +174,11 @@ private String getResponse(String url, String fieldParam) {
       return null;
     }
     JSONObject json = new JSONObject(responseBody.toString());
-    String res = json.getJSONArray(ITEM_PARAMETER).getJSONObject(0).get(fieldParam).toString();
+    JSONArray items = json.getJSONArray(ITEM_PARAMETER);
+    if (items == null || items.length() == 0) {
+      return null;
+    }
+    String res = items.getJSONObject(0).get(fieldParam).toString();
```
AaronLopes, line 181: nit: change to "result" or something more readable

anbenson:
Can you change the title to be a little more descriptive? I have to look up the issue each time :P

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
irisliu77: both ways look fine i think..?

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
kaylaleung: Unresolved?

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
kaylaleung: Uh, defer to Andrew and Joan?

But I still don't think you're safely accessing the fieldParam here?

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
irisliu77: uh alright, i just feel res is somewhat common

kaylaleung:
SO card error seems fixed, but actually, the original issue opened in #39 is still not completely addressed, because of lack of null check for links #65

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -173,10 +174,14 @@ private String getResponse(String url, String fieldParam) {
...
     JSONObject json = new JSONObject(responseBody.toString());
-    String res = json.getJSONArray(ITEM_PARAMETER).getJSONObject(0).get(fieldParam).toString();
-    if (response.getStatusLine().getStatusCode() != 200) {
-      return null;
+    if (json.has(ITEM_PARAMETER) && json.getJSONArray(ITEM_PARAMETER).length() > 0) {
+      JSONArray items = json.getJSONArray(ITEM_PARAMETER);
+      JSONObject obj = items.getJSONObject(0);
+      if (obj.has(fieldParam)) {
+        String result = obj.get(fieldParam).toString();
+        return result;
```
kaylaleung, line 182: can just directly `return obj.get(fieldParam).toString();` here

PR #56, opened 2020-07-21
Fix package path

kaylaleung:
Accidentally broke the test without updating the project with .external in #50

### DIFF
--- a/src/test/java/com/google/step/snippet/external/W3SchoolsClientTest.java
+++ b/src/test/java/com/google/step/snippet/external/W3SchoolsClientTest.java
@@ -1,10 +1,9 @@
-package com.google.step.snippet;
+package com.google.step.snippet.external;
 
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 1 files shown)


src/test/java/com/google/step/snippet/external/W3SchoolsClientTest.java
```diff
@@ -1,10 +1,9 @@
-package com.google.step.snippet;
+package com.google.step.snippet.external;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 
 import com.google.step.snippet.data.Card;
-import com.google.step.snippet.external.W3SchoolsClient;
```
joanchenx, line 7: Is removing this intentional?

kaylaleung: Yeah, it's no longer needed now that the package path is fixed

Issue #53, opened 2020-07-21
Searchbox positioning


AaronLopes:
css for the searchbox needs to be written such that it resembles the look in the mocks.

kaylaleung:
Do you mean homepage or search results page or both? Or like what part of the search box version of the mock?

AaronLopes:
Both, the mocks can be found under UI Mocks in the Snippet design doc.

PR #59, opened 2020-07-21
Add styling to Homepage and SRP searchbox view

kaylaleung:
Address #53 , change UI design to fit closer to design of UI mocks


src/main/webapp/css/home_style.css
```diff
@@ -0,0 +1,13 @@
+.heading {
+	display: flex;
+	flex-direction: column;
+	justify-content: flex-end;
+	align-items: center;
+	height: 50vh;
```
AaronLopes, line 6: style nit: when this is changed to 40, the searchbox looks a lot more centered.

irisliu77:
remove unrelated file in this pr

kaylaleung:
> remove unrelated file in this pr

There are no unrelated files? The new files are the icon images. And #56 should be merged into master first to resolve the .external thing

src/main/webapp/css/home_style.css
```diff
@@ -0,0 +1,13 @@
+.heading {
+	display: flex;
+	flex-direction: column;
+	justify-content: flex-end;
+	align-items: center;
+	height: 45vh;
```
joanchenx, line 6: The tabbing is doubled here I think.

src/main/webapp/css/style.css
```diff
@@ -22,3 +17,11 @@ h1 {
 .green {
   color: #0F9D58;
 }
+
+.logo {
+  text-decoration: none;
+}
+
+#___gcse_0 {
```
anbenson, line 25: Ideally we avoid using this CSS ID which could be changed by the CSE implementation. Perhaps target a div wrapper around the CSE searchbox?

src/main/webapp/css/style.css
```diff
@@ -22,3 +17,11 @@ h1 {
...
+  width: 60%;
```
anbenson, line 26: Adjusting to the parent width is nice, but you probably want to set a min-width and a max-width.

src/main/webapp/css/home_style.css
```diff
@@ -0,0 +1,13 @@
+.heading {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  align-items: center;
+  height: 45vh;
+}
+
+h1 {
+  text-align: center;
```
anbenson, line 10: This does nothing due to the way you do align-items: center, which forces the width of the h1 to be auto, which means centering doesn't make sense.

src/main/webapp/css/home_style.css
```diff
@@ -0,0 +1,13 @@
...
+  margin: 20px 20px 25px;
```
anbenson, line 11: I think you'll want the margin to be on the link, not the h1 - see how currently this margin is part of the link so the clickable area is pretty large.

src/main/webapp/css/srp_style.css
```diff
@@ -6,6 +6,17 @@
   flex-grow: 1;
 }
 
+.heading {
+  display: flex;
+  align-items: center;
+}
+
+h1 {
+  text-align: center;
```
anbenson, line 15: Ditto, and ditto for the margin again.

src/main/webapp/css/style.css
```diff
@@ -23,6 +18,12 @@ h1 {
   color: #0F9D58;
 }
 
+.searchbox {
+  width: 60%;
+  min-width: 400px;
+  max-width: 1000px;
```
anbenson, line 24: 1000px is actually really large. For comparison, the google.com searchbox ranges from 400px to 584px, and that's including the search and microphone buttons inside.

I'm actually not sure about the 60% anymore. Why waste 40% of the screen width on small screens?

src/main/webapp/WEB-INF/tags/base.tag
```diff
@@ -1,8 +1,10 @@
...
-		<span class="blue">S</span><span class="red">n</span><span class="yellow">i</span><span class="blue">p</span><span class="green">p</span><span class="red">e</span><span class="yellow">t</span>
-	</h1>
-	<div class="gcse-searchbox-only" data-resultsUrl="/search"></div>
-	<a class="auth-button" href="${authUrl}">${authLabel}</a>
+  <a class="logo" href="/">
+    <h1>
+      <span class="blue">S</span><span class="red">n</span><span class="yellow">i</span><span class="blue">p</span><span class="green">p</span><span class="red">e</span><span class="yellow">t</span>
+    </h1>
+  </a>
+  <div class="searchbox"><div class="gcse-searchbox-only" data-resultsUrl="/search"></div></div>
```
anbenson, line 8: Let's indent this as normal too, like

```html
<div class="searchbox">
  <div class="gcse-searchbox-only" ...></div>
</div>
```

src/main/webapp/css/srp_style.css
```diff
@@ -11,6 +11,15 @@
   align-items: center;
 }
 
+.logo {
+  margin: 0 20px;
+  text-decoration: none;
+}
+
```
anbenson, line 18: So the searchbox here looks off center with respect to the login button. This confused me for a long time, but I figured it out - the &lt;form&gt; and &lt;table&gt; inside the searchbox have some bottom margin, and removing it causes it to center correctly.

You might also consider aligning the login button's height to the search button's height, as it currently looks visually off.

You might also consider putting some margin between the search button and the login button, as they're currently stuck together...

kaylaleung: I moved the button outside flexbox because I couldn't get the position to fix relative to the top right corner... Unsure if this is the best way though.

src/main/webapp/css/home_style.css
```diff
@@ -0,0 +1,30 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  align-items: center;
+  height: 50vh;
+}
+
+h1 {
+  font: 400 60px Roboto, Arial, sans-serif;
+  margin: 0px 0px;
```
anbenson, line 11: Google style is to just do:

```css
margin: 0;
```

anbenson: (Also for srp style.)

src/main/webapp/css/srp_style.css
anbenson: Maybe not the way I'd do it, but this is fine.

src/main/webapp/css/srp_style.css
```diff
@@ -43,3 +54,17 @@
   margin-top: 0;
   display: block;
 }
+
+.auth-button {
+  position: absolute;
+  top: 35px;
+  right: 30px;
+  margin-bottom: 5px;
```
anbenson, line 62: Margin does nothing when you're positioned absolutely and are completely alone.

src/main/webapp/css/srp_style.css
```diff
@@ -43,3 +54,17 @@
...
+  font: 400 15px Roboto, Arial, sans-serif;
+  color: white;
+  background: #4d90fe;
+  border: 1px #3079ED solid;
```
anbenson, line 66: Consider moving out the common parts of the login button CSS to style.css.

src/main/webapp/css/style.css
```diff
@@ -23,14 +18,10 @@ h1 {
...
-.auth-button {
-  font: 400 15px Roboto, Arial, sans-serif;
-  color: white;
-  background: #4d90fe;
-  border: 1px #3079ED solid;
-  padding: 5px 10px;
-  border-radius: 4px;
-  text-decoration: none;
+.searchbox {
+  width: 100vh;
```
anbenson, line 22: If you just want "as large as my container lets me", do 100%. You don't actually mean 100vh - for example, on the SRP, this wouldn't leave any room for the logo, even if it were possible.

src/main/webapp/css/home_style.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.logo {
```
anbenson, line 14: nit: In your HTML, .logo comes before h1 so it makes more sense to me to have the rule come first as well.

Issue #44, opened 2020-07-20
Determine a solution to the HTML escaping issue


anbenson:
Currently HTML in Cards is not escaped. This brings a couple issues.
- This leaves the application open to Cross-Site Scripting attacks. If code scraped into the cards contains JavaScript, such untrusted code would automatically run on the user's computer.
- HTML example code is rendered as HTML instead of code that the user can see.

However it's not as simple as just escaping all HTML we see. Doing so shows the user unrendered HTML which can be confusing and hard to read. Stripping HTML in some cases might be a good idea.

irisliu77:
for stackoverflow, i don't see the need of escaping html, it's in the correct format #63

anbenson:
For posterity, here's what we decided in today's meeting (feel free to correct if I'm wrong):
1) Remove the 200 or 300 character limit in the cards. (Higher priority because doing so will avoid some of the merged-cards issues we're seeing.)
- Sanitize cards with jsoup.clean
- Give cards an initial width and height, and allow them to be scrolled horizontally and vertically

anbenson:
One last thing - for cards about HTML code, we might have to escape instead.

PR #63, opened 2020-07-21
remove code and only display the description in SO client

irisliu77:
also the code display looks better if it's display inline, i think this is changed in #27 #


src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -73,8 +69,8 @@ public Card search(String url) {
       return null;
     }
     // No description or code is allowed for StackOverflow.
-    String description = getDescription(answerBody);
-    String code = getCode(answerBody);
+    String description = getAnswerBody(answerId);
+    String code = null;
```
kaylaleung, line 73: Setting this to null leads to `<code></code>` being generated within the card on search.jsp.

There should be a null check included (probably within search.jsp itself, to make sure this empty code tag is not generated when `knowledgeCard.getCode() == null`)

Otherwise this new format looks way really good!

irisliu77: dealt with this in this pr but probably also need to be solve in #27 ?

kaylaleung: I think fixing here should be fine. The null check wasn't required before because we did the null check in the scraping clients themselves and never returned a new card if code was null.

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -110,35 +105,12 @@ private String getTitle(String questionId) {
...
-  /* Return the code parsed from answer body. */
-  private String getCode(String body) {
-    Document doc = Jsoup.parse(body);
-    // Combine all code in the answer body.
-    Elements codeHtml = doc.select(CODE_PARAMETER);
-    String code = "";
-    for (Element e : codeHtml) {
-      code += e.outerHtml();
+    String answerBody = getResponse(answerUrl, BODY_PARAMETER);
+    if (answerBody.length() > 300) {
```
kaylaleung, line 109: getResponse can return null. Add a null check here before trying to access length()

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -73,8 +68,8 @@ public Card search(String url) {
       return null;
     }
     // No description or code is allowed for StackOverflow.
-    String description = getDescription(answerBody);
-    String code = getCode(answerBody);
+    String description = getAnswerBody(answerId);
```
kaylaleung, line 71: Oh wait, hmm if the description is null or empty, no card should be created here right?

irisliu77: oh wrong method called here, already got the answerbody above

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -22,7 +22,9 @@
             <a href="${knowledgeCard.getLink()}" title="card-source"class="card-title">
               ${knowledgeCard.getTitle()}
             </a>
-            <code>${knowledgeCard.getCode()}</code>
+            <c:if test="${knowledgeCard.getCode() != null}">
+              <code>${knowledgeCard.getCode()}</code>
+            </c:if>
             <span class="card-description">Description:</span>
```
kaylaleung, line 28: Turns out removing this `<code></code>` actually makes the description wrap with the title since it's a span and not a div

irisliu77: that happens sometime, but for me it make more sense to change the tag for description to <div>

anbenson:
Can you fix the title of this PR so it's clear this is only for the StackOverflow parser?

I also notice this doesn't do anything with the HTML, so we're still going to hit the issue where cutting off at 300 characters may cut off in the middle of a HTML tag, causing invalid HTML. Also XSS attacks and all the fun stuff.

Ideally the fix here is compliant with whatever was decided about HTML escaping/stripping/etc.

kaylaleung:
The description cutoff shouldn't affect or break any HTML code right now cause no escaping is being applied to SO cards. But unclear what is happening with escaping. Maybe description cutoff can just be moved to be a different PR that waits for escaping to be addressed and resolved.

anbenson:
So as per our latest discussion in today's meeting (documented in #44), and per the comment in #74, I'm expecting this PR to change so that it removes the max length cutoff (and the ...) and sanitizes HTML. Let me know if this differs from what you were expecting.

irisliu77:
some commits here are redundant because i added test for SO to this branch accidentally

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -69,18 +63,17 @@ public Card search(String url) {
...
+    String code = null;
     return new Card(title, code, url, description);
   }
 
   /* Get the question id of passed in URL. */
-  private String getQuestionId(String url) {
+  String getQuestionId(String url) {
     URI uri;
     try {
       uri = new URI(url);
```
kaylaleung, line 79: new URI can also throw a NullPointerException when url is null, although i guess URL should never be null if its gotten this far.

irisliu77: url will not be null. if it's null, the search servlet will skip searching method for this client.

anbenson:
(Requesting review from Aaron)

Can you leave the removal of privates to your PR that adds tests? I'd rather not conflate changes across PRs.

Will you be adding sanitizing HTML in this PR too (as per my last comment), or will that come in a new PR?

irisliu77:
> (Requesting review from Aaron)
> 
> Can you leave the removal of privates to your PR that adds tests? I'd rather not conflate changes across PRs.
> 
> Will you be adding sanitizing HTML in this PR too (as per my last comment), or will that come in a new PR?

I think sanitizing is done in this pr

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -71,13 +66,12 @@ public Card search(String url) {
...
-    if (title == null || answerBody == null) {
+    String description = getAnswerBody(answerId);
+    if (title == null || description == null) {
       return null;
     }
-    // No description or code is allowed for StackOverflow.
-    String description = getDescription(answerBody);
-    String code = getCode(answerBody);
+    // Code is intentionally set to null for better display.
+    String code = null;
```
AaronLopes, line 74: why not just set the param to null in the return statement?

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -71,13 +66,12 @@ public Card search(String url) {
...
     return new Card(title, code, url, description, SOURCE_NAME, ICON_LINK);
```
AaronLopes, line 75: are the other fields (title/description) going to be cleaned here?

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
joanchenx: Yeah, I think you can just set to null in the return statement. If you do that, can you do add a comment so it's clear which field is null? 

`return new Card(title, /* code= */ null, url, description, SOURCE_NAME, ICON_LINK);`

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
irisliu77: i think the title is a string that don't need to be cleaned. Also, for description, i cannot think of a certain tag that need to be cleaned at this moment

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -91,6 +84,9 @@ private String getQuestionId(String url) {
     }
     // Parse the URL to get the question id.
     String[] segments = uri.getPath().split("/");
+    if (segments.length <= ID_INDEX || !segments[QUESTION_INDEX].equals(QUESTION_PARAMETER)) {
```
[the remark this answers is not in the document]
joanchenx: Also for `QUESTION_INDEX`, I feel like it might be better if it was named something like `QUESTION_PATH_INDEX` or something that better indicates what it represents? Not sure if you have a better name?

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -91,6 +84,9 @@ private String getQuestionId(String url) {
     }
     // Parse the URL to get the question id.
     String[] segments = uri.getPath().split("/");
+    if (segments.length <= ID_INDEX || !segments[QUESTION_PATH_INDEX].equals(QUESTION_ID_INDEX)) {
```
joanchenx, line 87: Ah sorry, for the confusion, I was thinking replacing `ID_INDEX` with `QUESTION_ID_INDEX` and keeping `QUESTION_PARAMETER` so:

`if (segments.length <= QUESTION_ID_INDEX || !segments[QUESTION_PATH_INDEX].equals(QUESTION_PARAMETER))`

src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -71,14 +66,12 @@ public Card search(String url) {
...
-    if (title == null || answerBody == null) {
+    String description = getAnswerBody(answerId);
+    if (title == null || description == null) {
       return null;
     }
-    // No description or code is allowed for StackOverflow.
-    String description = getDescription(answerBody);
-    String code = getCode(answerBody);
-    return new Card(title, code, url, description, SOURCE_NAME, ICON_LINK);
+    // Code is intentionally set to null for better display.
```
anbenson, line 73: I don't understand what this comment means...

PR #64, opened 2020-07-21
Implement Feedback Servlet

kaylaleung:
Notes:
- Storring additional upvote and downvote information for potential future use?
- Also I believe datastore stores numerical fields as longs, hence the long casts.


src/main/java/com/google/step/snippet/external/W3SchoolsClient.java
```diff
@@ -52,6 +53,15 @@ public Card search(String w3Link) {
...
+    /* Retrieve feedback, if stored feedback exists */
+    long upvote = 0;
+    long downvote = 0;
+    Entity feedback = getFeedback(w3Link);
+    if (feedback != null) {
+      upvote = (long) feedback.getProperty(UP);
+      downvote = (long) feedback.getProperty(DOWN);
+    }
+    return new Card(title, code, w3Link, description, upvote, downvote);
   }
```
irisliu77, line 66: this can be a static method in the client

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,77 @@
...
+  private static final String URL = "url";
+  private static final String FEEDBACK = "feedback";
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Entity feedbackEntity = null;
+    String url = request.getParameter(URL);
+
+    if (url == null) {
+      System.out.println("Invalid Card");
```
irisliu77, line 28: remove print statement

src/main/java/com/google/step/snippet/external/Client.java
```diff
@@ -1,9 +1,43 @@
 package com.google.step.snippet.external;
 
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.FilterOperator;
 import com.google.step.snippet.data.Card;
 
 public interface Client {
```
anbenson, line 10: This is no longer an interface, but an abstract class. For example, `getFeedback` is not a method you're publicizing as part of your interface, but a protected method you want subclasses to have access to.

src/main/java/com/google/step/snippet/data/Card.java
```diff
@@ -5,12 +5,17 @@
...
+  private final long downvote;
 
-  public Card(String title, String code, String link, String description) {
+  public Card(
+      String title, String code, String link, String description, long upvote, long downvote) {
     this.title = title;
     this.code = code;
     this.link = link;
     this.description = description;
+    this.upvote = upvote;
```
anbenson, line 17: I recall you mentioning previously that both upvotes and downvotes (these should probably be plural? singular makes it sound like a boolean) are kept in the backend for future unknown use. But the frontend only cares about the difference. Let's only give the frontend exactly as much information as they need, so only the difference?

anbenson: This has ramifications on a lot of your PR - for example, Client.java only needs one method that tells you what the feedback value (difference) is, not a method to get back an entity (which is an implementation detail), one for upvotes, and one for downvotes.

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,76 @@
...
+  private static final String URL = "url";
+  private static final String FEEDBACK = "feedback";
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Entity feedbackEntity = null;
+    String url = request.getParameter(URL);
+
+    if (url == null) {
+      response.sendRedirect(request.getHeader("Referer"));
```
anbenson, line 28: It's bad practice to depend on user-supplied information, like request headers, without validating it. But I'm not sure what you're trying to do here - isn't this an async post request? I don't think you want to do any redirects.

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /* Query for card feedback by card URL */
+    Query.FilterPredicate filter = new Query.FilterPredicate(URL, FilterOperator.EQUAL, url);
+    Query query = new Query(FEEDBACK).setFilter(filter);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    feedbackEntity = datastore.prepare(query).asSingleEntity();
+    long totalVotes = 0;
+
+    if (feedbackEntity == null) {
```
anbenson, line 38: Feels like a lot of duplicated code, and I feel like you could condense this into maybe 1/4 as many lines. The general flow I would expect is

1. If there's no entity, make one with default values.
2. Figure out from the request what the change in upvotes/downvotes will be.
3. Make the changes.
4. Find the new value of upvotes/downvotes for your total.

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,76 @@
...
+      }
+      /* Update with downvote count */
+      else if (request.getParameter(DOWN) != null) {
+        totalVotes =
+            (long) feedbackEntity.getProperty(UP) - (long) feedbackEntity.getProperty(DOWN) - 1;
+        feedbackEntity.setProperty(DOWN, (long) feedbackEntity.getProperty(DOWN) + 1);
+      }
+    }
+    /* Convert vote count to response string */
+    String totalStr;
```
anbenson, line 66: Are you sure Long.toString doesn't do this for you?

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
kaylaleung: See #76 .

The request headers are being set by the POST request. I'm not sure what you mean by validating here beyond making null checks.

irisliu77:
if we only wants total votes in the frontend, i think we can only keep one totalvotes field

kaylaleung:
> if we only wants total votes in the frontend, i think we can only keep one totalvotes field

Yeah, for this PR, having multiple fields to track upvote and downvote isn't absolutely necessary, but the idea is that the list of upvotes and downvotes will be helpful in the future if we're tracking the list of people who have already upvoted vs downvoted and prevent toggling multiple times

PR #67, opened 2020-07-22
Card bleeding

irisliu77:


src/main/webapp/css/srp_style.css
```diff
@@ -21,6 +21,7 @@
 
 .card .card-title {
   font: 300 20px Roboto, Arial, sans-serif;
+  word-wrap: break-word;
 }
```
kaylaleung, line 25: I don't think this does anything? At least it's still showing up as the code and title wrapping together instead of being on separate lines.

irisliu77: this pr deal with long title that bleeds out the card

PR #69, opened 2020-07-22
Remove title attribute

### LINKED ISSUE #68: Missing space and incorrect use of title attribute in search.jsp
https://github.com/googleinterns/step1-2020/blob/master/src/main/webapp/WEB-INF/templates/search.jsp#L21

There is no space between class and title.

Additionally, title becomes the hover text of the link - it should not be something like "card-source", but actual informative text about the link, if there is a title at all.

AaronLopes:
Fixes #68


anbenson:
Not sure what's going on here, but you might need to retrigger the check by amending no changes to your last commit and force-pushing, or something like that.

PR #73, opened 2020-07-22
Fixed Null link check

irisliu77:


src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -100,7 +100,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     for (Client client : clients) {
       String link = getLink(client.getCseId(), query);
       if (link == null) {
```
anbenson, line 102: I would prefer

```java
for (Client client : clients) {
  if (link != null) {
    Card card = client.search(link);
    if (card != null) {
      allCards.add(card);
    }
  }
}
```

This way the checks are consistent instead of first checking for null to continue, then checking for not null to do work.

PR #74, opened 2020-07-22
Sanitize and escape html

### LINKED ISSUE #44: Determine a solution to the HTML escaping issue

AaronLopes:
Fixes #44 (review test files in #78)


anbenson:
1) My understanding is that in general we will jsoup.clean (sanitize) but not escape. If you escape, you're going to see &lt;code&gt;&lt;p&gt;some text&lt;/p&gt;&lt;code&gt; everywhere.
2) Why not the description field?
3) I notice the StackOverflow client is excluded here. Are there separate changes for those in another PR, or what's the status of that?

AaronLopes:
> 3\. I notice the StackOverflow client is excluded here. Are there separate changes for those in another PR, or what's the status of that?

1. I don't remember there ever being nested tags in either W3/GeeksForGeeks in a non-html context (on queries tested so far/sifting through the site). 
2. Same here regarding tags in the descriptions.
3. It was my understanding that escaping/stripping regarding SO was being done in #63, as per the latest comments.

anbenson:
It's safe to clean/escape things that aren't HTML too. Whether you do so or not should depend on your intent - is the string you're considering untrusted input? In this case, all fields of the Card come from the wild (scraped websites), so they're untrusted, so you'll need to clean or escape everything, regardless of whether you expect there to be HTML or not. If you don't do so, then it's an avenue for a XSS attack. (Maybe you've never seen there being tags in the W3Schools description field, but if that causes you not to sanitize it, then a malicious actor with control of the W3Schools website could carry out an XSS attack on you.)

For SO - noted.

Still confused on why escaping is being done, though - I thought we would only do that if we know the field contains HTML code (as a hack to allow users to search for HTML code).

AaronLopes:
> It's safe to clean/escape things that aren't HTML too. Whether you do so or not should depend on your intent - is the string you're considering untrusted input? In this case, all fields of the Card come from the wild (scraped websites), so they're untrusted, so you'll need to clean or escape everything, regardless of whether you expect there to be HTML or not. If you don't do so, then it's an avenue for a XSS attack. (Maybe you've never seen there being tags in the W3Schools description field, but if that causes you not to sanitize it, then a malicious actor with control of the W3Schools website could carry out an XSS attack on you.)
> 
> For SO - noted.
> 
> Still confused on why escaping is being done, though - I thought we would only do that if we know the field contains HTML code (as a hack to allow users to search for HTML code).

Gotcha. 

After our meeting, I was under the impression that detecting html before escaping was irrelevant and implies the use of regex, which we wanted to rule out (wasn't aware of that hack though, which is a good idea).

anbenson:
I hope I wasn't misleading. Using regex to parse HTML is indeed imperfect, and I recommended not using it to determine whether to sanitize because 1) it's imperfect and 2) the regex engine is less efficient than just sanitizing it anyway, which is a simple (ish) O(n) sweep.

Yes, the hack I think we're doing by escaping HTML when we detect there may be HTML code (the simplest way to do this is probably just checking if the query contains "html"?) does open us up to XSS attacks and is Not Great. But the hope is to protect ourselves in most cases (e.g. when the query does not contain "html") and make sure the app is still usable for HTML queries.

kaylaleung:
@ iris and I when ready for review!

src/main/java/com/google/step/snippet/external/W3SchoolsClient.java
```diff
@@ -49,9 +51,18 @@ public Card search(String w3Link) {
...
+            Jsoup.clean(titles.first().text(), Whitelist.basicWithImages()));
+    String description =
+        StringEscapeUtils.escapeHtml4(
+            Jsoup.clean(descriptions.first().text(), Whitelist.basicWithImages()));
+    String code =
+        Jsoup.clean(
+            snippets.first().getElementsByClass(CODE_CLASS).text(), Whitelist.basicWithImages());
+    if (query.contains("html")) {
+      code = StringEscapeUtils.escapeHtml4(code);
+    }
```
irisliu77, line 65: duplicate code here in two clients, can move to either Client interface or Card class

src/main/java/com/google/step/snippet/external/W3SchoolsClient.java
```diff
@@ -49,9 +51,18 @@ public Card search(String w3Link) {
...
+        StringEscapeUtils.escapeHtml4(
+            Jsoup.clean(titles.first().text(), Whitelist.basicWithImages()));
+    String description =
+        StringEscapeUtils.escapeHtml4(
+            Jsoup.clean(descriptions.first().text(), Whitelist.basicWithImages()));
+    String code =
+        Jsoup.clean(
+            snippets.first().getElementsByClass(CODE_CLASS).text(), Whitelist.basicWithImages());
+    if (query.contains("html")) {
+      code = StringEscapeUtils.escapeHtml4(code);
```
kaylaleung, line 64: I think at least for W3Schools, checking if the URL also contains html is a more reliable check way to check if escaping is needed. Up to you if you want to do the same thing for GeeksForGeeks

kaylaleung: ex: just searching "images" yields unescaped html code from W3

AaronLopes: Yeah, was thinking of other filters we might want to add. There are other keywords (eg. svg, icons, etc.) that would also return html code that isn't escaped.

src/main/java/com/google/step/snippet/external/W3SchoolsClient.java
```diff
@@ -60,9 +63,18 @@ public Card search(String w3Link, String query) {
     String code =
         Jsoup.clean(
             snippets.first().getElementsByClass(CODE_CLASS).text(), Whitelist.basicWithImages());
```
kaylaleung, line 65: A concerning issue here, that I don't have all the answers to cause I don't really understand Jsoup.clean that well, but here goes... For the ["css style"](https://www.w3schools.com/html/html_css.asp) query

Before cleaning, the code retrieved is of the following form, which matches with what is on the website.
```
<h1 style="color:blue;">A Blue Heading</h1> <p style="color:red;">A red paragraph.</p>
```
And since it should be escaped, the expected escaped form is
```
&lt;h1 style=&quot;color:blue;&quot;&gt;A Blue Heading&lt;/h1&gt; &lt;p style=&quot;color:red;&quot;&gt;A red paragraph.&lt;/p&gt;
```
However, after Jsoup.clean, the non escaped code is in fact
```
A Blue Heading 
<p>A red paragraph.</p>
```
Which, when escaping is then applied to it, turns into
```
A Blue Heading 
&lt;p&gt;A red paragraph.&lt;/p&gt;
```
["css flexbox"](https://www.w3schools.com/css/css3_flexbox.asp) query is another test. It should be producing 
```
<div class="flex-container"> <div>1</div> <div>2</div> <div>3</div> </div>
```
But after clean, instead, it is just `1 2 3`

Seems like HTML escaping is doing its job, but Jsoup.clean is stripping away tags that it shouldn't be.

anbenson: https://jsoup.org/apidocs/org/jsoup/safety/Whitelist.html#basicWithImages()

Yeah, clean removes all tags except a whitelist, and that whitelist does not contain div.

AaronLopes: Ah, my mistake. I'll change the whitelist param.

src/main/java/com/google/step/snippet/external/GeeksForGeeksClient.java
```diff
@@ -60,9 +63,18 @@ public Card search(String geeksLink, String query) {
...
   }
+
+  private boolean doEscape(String possibleHtml) {
+    for (String filterWord : escapeFilters) {
+      if (possibleHtml.contains(filterWord)) {
+        return true;
+      }
+    }
+    return false;
+  }
```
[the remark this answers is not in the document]
AaronLopes: I thought it best not to do this because the escaping in StackOverflow wouldn't require this method, as its being done separately in #63

anbenson:
This PR contains the GeeksForGeeks test which doesn't seem intentional.

My understanding was that you would only escape the code if the query contained 'html' or whatever heuristic is being used, but I see that most fields are unconditionally both sanitized and escaped.

src/test/java/com/google/step/snippet/external/W3SchoolsClientTest.java
```diff
@@ -15,19 +15,20 @@
 
   @Test
   public void htmlCodeCard() {
-    Card actual = client.search("https://www.w3schools.com/tags/tag_img.asp");
+    Card actual = client.search("https://www.w3schools.com/tags/tag_img.asp", "html img");
     Card expected =
         new Card(
-            "HTML <img> Tag",
-            "<img src=\"img_girl.jpg\" alt=\"Girl in a jacket\" width=\"500\" height=\"600\">",
+            "HTML \n&lt;img&gt; Tag",
```
kaylaleung, line 21: So the div tag removal thing seems resolved, but Jsoup.clean is still stripping away class settings, and style settings in a tag. 

In the htmlCodeCard test, it's stripping away the `src=` field, in cssEscaping `style=` is being stripped, and in divEscaping `class=`.

I wrote up some tests that should help check this. 
```
 @Test
  public void htmlCodeCard() {
    Card actual = client.search("https://www.w3schools.com/tags/tag_img.asp", "html img");
    Card expected =
        new Card(
            "HTML \n&lt;img&gt; Tag",
            "&lt;img src=&quot;img_girl.jpg&quot; alt=&quot;Girl in a jacket&quot; width=&quot;500&quot; height=&quot;600&quot;&gt;",
            "https://www.w3schools.com/tags/tag_img.asp",
            "How to insert an image:");
    assertEquals(expected, actual);
  }
```
```
  @Test
  public void cssEscaping() {
    Card actual = client.search("https://www.w3schools.com/html/html_css.asp", "css style");
    Card expected =
        new Card(
            "HTML Styles - CSS",
            "&lt;h1 style=&quot;color:blue;&quot;&gt;A Blue Heading&lt;/h1&gt; &lt;p style=&quot;color:red;&quot;&gt;A red paragraph.&lt;/p&gt;",
            "https://www.w3schools.com/html/html_css.asp",
            "CSS stands for Cascading Style Sheets.");
    assertEquals(expected, actual);
  }
```
```
  @Test
  public void divEscaping() {
    Card actual = client.search("https://www.w3schools.com/css/css3_flexbox.asp", "css flexbox");
    Card expected =
        new Card(
            "CSS Flexbox",
            "&lt;div class=&quot;flex-container&quot;&gt; &lt;div&gt;1&lt;/div&gt; &lt;div&gt;2&lt;/div&gt; &lt;div&gt;3&lt;/div&gt; &lt;/div&gt;",
            "https://www.w3schools.com/css/css3_flexbox.asp",
            "Try it Yourself &raquo;");
    assertEquals(expected, actual);
  }
}
```

AaronLopes: I was aware of the stripping being done to these fields, as (I believe) clean/Whitelist has its own heuristics for what constitutes a "safe" param for these fields. Example: for the src field, only http and https links are allowed. However, I know that our definition of "safe" might be different for the sake of preserving fields like this. Thoughts?

src/main/java/com/google/step/snippet/external/GeeksForGeeksClient.java
```diff
@@ -54,6 +59,27 @@ public Card search(String geeksLink) {
     String title = titles.first().text();
     String description = descriptions.first().text();
     String code = snippets.first().getElementsByClass(CODE_CLASS).text();
+    if (containsEscape(query.toLowerCase())
+        || containsEscape(geeksLink)
+        || containsEscape(title.toLowerCase())
+        || containsEscape(description.toLowerCase())) {
+      title = StringEscapeUtils.escapeHtml4(title);
+      description = StringEscapeUtils.escapeHtml4(description);
```
kaylaleung, line 67: Maybe we should only not clean and escape the code section and stick with cleaning title and description regardless. I doubt there will be any valuable HTML code that we need to display in the description or title.

kaylaleung:
Are you planning to add the other tests in a different PR?

src/main/java/com/google/step/snippet/external/GeeksForGeeksClient.java
anbenson: Agreed, except for the SO client which after Iris' PR will put code in the description

AaronLopes:
> Are you planning to add the other tests in a different PR?

The GeeksForGeeks tests? They're in #78

kaylaleung:
> > Are you planning to add the other tests in a different PR?
> 
> The GeeksForGeeks tests? They're in #78

I meant the escaping tests for W3Schools

AaronLopes:
> > > Are you planning to add the other tests in a different PR?
> > 
> > 
> > The GeeksForGeeks tests? They're in #78
> 
> I meant the escaping tests for W3Schools

Yes, these should be in a separate PR.

joanchenx:
> > > > Are you planning to add the other tests in a different PR?
> > > 
> > > 
> > > The GeeksForGeeks tests? They're in #78
> > 
> > 
> > I meant the escaping tests for W3Schools
> 
> Yes, these should be in a separate PR.

Can you create an issue for this?

src/main/java/com/google/step/snippet/external/GeeksForGeeksClient.java
```diff
@@ -51,9 +56,28 @@ public Card search(String geeksLink) {
...
+      code = StringEscapeUtils.escapeHtml4(code);
+    } else {
+      code = Jsoup.clean(code, Whitelist.relaxed().preserveRelativeLinks(true));
```
anbenson, line 70: Why do you want to preserve relative links? Won't any such links fail, since the resource won't be hosted on your app?

src/test/java/com/google/step/snippet/external/W3SchoolsClientTest.java
```diff
@@ -44,19 +45,19 @@ public void jsonCodeCard() {
...
   public void partiallyFilledCard() {
-    Card actual = client.search("https://www.w3schools.com/html/html_quiz.asp");
+    Card actual = client.search("https://www.w3schools.com/html/html_quiz.asp", "html quiz");
     assertNull(actual);
   }
 
   @Test
   public void nonexistentLink() {
-    Card actual = client.search("https://www.w3schools.com/css/html.asp");
+    Card actual = client.search("https://www.w3schools.com/css/html.asp", null);
```
anbenson, line 54: Let's use `""` instead of `null`. Your code implicitly assumes the query is non-null since it calls `toLowerCase()` on it.

src/main/java/com/google/step/snippet/external/GeeksForGeeksClient.java
```diff
@@ -51,9 +56,28 @@ public Card search(String geeksLink) {
...
+    }
     return new Card(title, code, geeksLink, description, SOURCE_NAME, ICON_LINK);
   }
+
+  private boolean containsEscape(String possibleHtml) {
```
anbenson, line 75: I'm going to expect a future PR to move this to the Client abstract class (in this PR, it's still an interface).

src/main/java/com/google/step/snippet/external/GeeksForGeeksClient.java
AaronLopes: Sure, but I thought preserving these links might provide useful context in an answer for an html related question.

AaronLopes: Old code from when I was cleaning all displayed code, I thought preserving links might help provided context to html related questions.

PR #76, opened 2020-07-23
Add Feedback Voting UI 

kaylaleung:
- Add votes field to card class. 
- Change client from interface to abstract class. 
- Add styling to search.jsp for toggle up, toggle down, and count label. 
- Add javascript code to sent POST request anytime toggle buttons are clicked


src/main/webapp/js/script.js
```diff
@@ -0,0 +1,29 @@
...
+      renderVote(card, down.value, 'downvote');
+    });
+  }
+});
+
+async function renderVote(card, url, action) {
+  const total = card.getElementsByClassName('total-votes')[0];
+  total.innerHTML = await updateVote(url, action);
+}
+
```
irisliu77, line 19: i think you mentioned that you want to make t numbers longer than 3 digits to be display in a shorter way (1.0k)

kaylaleung: based on my searching, I don't think there's one clean way to do this and it will likely require some other js library. Probably better to make it a separate issue to be addressed in a future PR.

irisliu77:
i don't know whether it's because i have serious astigmatism, it's hard to see the green button after upvote. Maybe change the color if it's not my eyes' problem....

src/main/webapp/css/srp_style.css
```diff
@@ -37,33 +37,51 @@ h1 {
...
+}
+
+.vote button {
+  background: none;
+  border: none;
+  outline: none;
+}
+
+.total-votes {
+  padding: 1px 6px;
```
AaronLopes, line 81: this is (probably) bad practice, but adding ```margin-right: 3px``` centers total votes a bit better.

kaylaleung:
> overall lgtm, assuming the js functions work with the servlet

You should be able to double check the js works from the vote-auth branch, but heh yes

src/main/webapp/css/srp_style.css
kaylaleung: Changed around the centering with flexbox, seems to be better now

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -17,18 +19,29 @@
...
-              </a>
-            </header>
-            <c:if test="${knowledgeCard.getCode() != null}">
-              <code>${knowledgeCard.getCode()}</code>
-            </c:if>
-            <div class="card-description">Description:</div>
-            <p>${knowledgeCard.getDescription()}</p>
+          <div class="card" url="${knowledgeCard.getLink()}">
+            <div class="vote">
+              <button class="upvote">
```
anbenson, line 24: Can these buttons set the css cursor property?

src/main/java/com/google/step/snippet/data/Card.java
```diff
@@ -6,14 +6,22 @@
   private final String link;
   private final String description;
   private final String source;
+  private final String votes;
```
anbenson, line 9: Why string? This is a number.

src/test/java/com/google/step/snippet/external/GeeksForGeeksClientTest.java
```diff
@@ -11,7 +11,12 @@
 @RunWith(JUnit4.class)
 public final class GeeksForGeeksClientTest {
 
-  private final GeeksForGeeksClient client = new GeeksForGeeksClient("CSE_ID");
+  private final GeeksForGeeksClient client =
+      new GeeksForGeeksClient("CSE_ID") {
+        public String getVotes(String url) {
```
anbenson, line 16: Why public? But why is this necessary at all? This is the same implementation as in Client.java.

src/main/webapp/js/script.js
```diff
@@ -0,0 +1,61 @@
...
+    up.addEventListener('click', function() {
+      renderVote(card, 'upvote');
+    });
+    down.addEventListener('click', function() {
+      renderVote(card, 'downvote');
+    });
+  }
+});
+
+function initVote(cards) {
```
anbenson, line 16: This shouldn't be necessary. Can't this be done on initial render? But fine, I'll okay it.

src/main/webapp/js/script.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+    if (json.toggleDownvote === 'active') {
+      down.style.color = 'red';
+    } else {
+      down.style.color = 'black';
+    }
+  }
+}
+
+async function updateVote(url, action) {
```
anbenson, line 52: nit: put this before toggleButtons

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -8,6 +8,8 @@
     <t:head/>
     <link href="css/srp_style.css" rel="stylesheet">
     <title>Snippet</title>
+    <script type="text/javascript" src="${pageContext.request.contextPath}js/script.js"></script>
```
anbenson, line 11: Just js/script.js is fine.

src/test/java/com/google/step/snippet/external/GeeksForGeeksClientTest.java
kaylaleung: It won't be the same as the implementation in Client.java in the next PR.

kaylaleung: Also, this cant be public because the override cant reduce the visibility of the method inside Client.java

joanchenx: Can this be done in the next PR instead of this one??

src/main/webapp/js/script.js
kaylaleung: In my first implementation of this color loading, I did all the rendering, color included in SearchServlet inside the card that is returned but you mentioned that server should avoid being responsible for anything UI styling related.

anbenson: I meant that the API (the json returned by GET/POST requests) should not contain specific UI styling.

It's a bit confusing, but the server-rendered template (the JSP template) is a client of your server. It receives the cards from the servlet, and figures out how to render them in HTML, even though technically this code is part of the server.

The reason I'm saying it's fine is because you would need to add a new field to Card.java, and have both JSP code and JS code to change the color, and that's more work than I'm willing to ask you to do.

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -17,17 +19,28 @@
...
-              <img class="icon" src="${knowledgeCard.getIcon()}" alt="${knowledgeCard.getSource()} site icon" width="25px" height="25px">
-              <a href="${knowledgeCard.getLink()}" class="card-title">
-                ${knowledgeCard.getTitle()}
-              </a>
-            </header>
-            <c:if test="${knowledgeCard.getCode() != null}">
-              <pre><code>${knowledgeCard.getCode()}</code></pre>
-            </c:if>
-            <p>${knowledgeCard.getDescription()}</p>
+          <div class="card" url="${knowledgeCard.getLink()}">
```
anbenson, line 22: url is not a valid HTML attribute. What you're looking for is https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

src/main/webapp/css/srp_style.css
```diff
@@ -98,3 +102,21 @@ pre::-webkit-scrollbar {
...
+}
+
+.vote button {
+  cursor: pointer;
+  background: none;
+  border: none;
+}
+
+.total-votes {
+  font-size: 15px;
```
anbenson, line 121: If you remove this, you can get rid of the span and so forth in the HTML, making it simpler. The font will then be 16px, making it consistent with the header.

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -17,17 +19,28 @@
...
+            <div class="vote">
+              <button class="upvote">
+                <i class="fas fa-chevron-up"></i>
+              </button>
+              <span class="total-votes">${knowledgeCard.getVotes()}</span>
+              <button class="downvote">
+                <i class="fas fa-chevron-down"></i>
+              </button>
+            </div>
+            <div class="content">
```
anbenson, line 32: During this refactoring, the padding for the card got moved outside of the card content, so now the scrollbar appears over the card content. Can you move the padding back inside?

src/main/webapp/css/srp_style.css
kaylaleung: I still need to get access to total votes in the script.js file though so I don't think I can get rid of the span completely?

src/main/webapp/css/srp_style.css
```diff
@@ -98,3 +102,17 @@ pre::-webkit-scrollbar {
   margin-right: 5px;
   vertical-align: bottom;
 }
+
+.vote {
```
anbenson, line 106: Can you put this above  .content so the CSS order somewhat follows the order of appearance?

src/main/webapp/css/srp_style.css
```diff
@@ -37,46 +37,50 @@ h1 {
...
-  padding: 24px;
   box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 3px 1px -2px rgba(0, 0, 0, 0.12), 0 1px 5px 0 rgba(0, 0, 0, 0.2);
   transition: box-shadow .25s;
   border-radius: 2px;
   width: 400px;
   max-height: 260px;
+  display: flex;
+}
+
+.content {
```
anbenson, line 48: .card .content?

src/main/webapp/js/script.js
```diff
@@ -0,0 +1,62 @@
...
+    },
+  });
+  return response.json();
+}
+
+function toggleButtons(card, json) {
+  const total = card.getElementsByClassName('total-votes')[0];
+  if (Object.keys(json).length !== 0) {
+    if (json.totalvotes != null) {
+      total.innerHTML = json.totalvotes;
```
anbenson, line 46: Can you use innerText or something? Use of innerHTML is associated with security vulnerabilities.

PR #78, opened 2020-07-24
GeeksForGeeks tests

AaronLopes:
(review html escape changes in #74)


irisliu77:
good for now, but seems a lot of conflict will occur in the future

Issue #75, opened 2020-07-23
Searchbox suggestions


anbenson:
How difficult would it be to get suggestions to appear below the searchbox? I'm wondering if there's just a checkbox you need to toggle or something in the CSE dashboard?

I know the first thing our team would look for during a presentation is the searchbox suggestions.

kaylaleung:
See #58

Issue #45, opened 2020-07-20
Word wrapping in code makes code hard to read


anbenson:
The code in Cards are currently rendered to wrap lines. This can sometimes be hard to read since code isn't supposed to be word-wrapped.

An example, although not the best example: search for "how to get address of pointer in C", look at the GeeksForGeeks card

kaylaleung:
For GeekForGeeks in particular, a quick examination at the structure of their site HTML code suggests that each line can be parsed out separately rather than just .text() with jSoup because on the GFG site, each code block is actually composed of multiple div classes for each line of code represented by`<div class="line numberX indexX"`></div>

PR #79, opened 2020-07-24
add x and y scroll, autocomplete to searchbox

### LINKED ISSUE #55: Code snippet "grey box"
Add a grey box surrounding the code snippets, would highlight code and provide better separation between description and snippets.

AaronLopes:
Fixes #55, #75, #45.

### DIFF
--- a/src/main/webapp/WEB-INF/tags/base.tag
+++ b/src/main/webapp/WEB-INF/tags/base.tag
@@ -3,6 +3,6 @@
 	<h1>
 		<span class="blue">S</span><span class="red">n</span><span class="yellow">i</span><span class="blue">p</span><span class="green">p</span><span class="red">e</span><span class="yellow">t</span>
 	</h1>
-	<div class="gcse-searchbox-only" data-resultsUrl="/search"></div>
+	<div class="gcse-searchbox-only" enableAutoComplete="true" data-resultsUrl="/search"></div>
 	<a class="auth-button" href="${authUrl}">${authLabel}</a>
 </div>
--- a/src/main/webapp/WEB-INF/templates/search.jsp
+++ b/src/main/webapp/WEB-INF/templates/search.jsp
@@ -21,7 +21,7 @@
             <a href="${knowledgeCard.getLink()}" title="card-source"class="card-title">
               ${knowledgeCard.getTitle()}
             </a>
-            <code>${knowledgeCard.getCode()}</code>
+            <pre><code>${knowledgeCard.getCode()}</code></pre>
             <span class="card-description">Description:</span>
             <p>${knowledgeCard.getDescription()}</p>
           </div>
--- a/src/main/webapp/css/srp_style.css
+++ b/src/main/webapp/css/srp_style.css
@@ -22,6 +22,19 @@
   transition: box-shadow .25s;
   border-radius: 2px;
   width: 400px;
+  height: 260px;
+  overflow: scroll;
+}
+
+.card::-webkit-scrollbar {
+  width: 5px;
+  height: 5px;
+  background-color: WHITE;
+}
+
+pre::-webkit-scrollbar-thumb {
+  color: grey;
+  border-radius: 10px;
 }
 
 .card .card-title {
@@ -42,4 +55,23 @@
 .card p {
   margin-top: 0;
   display: block;
+}
+
+pre {
+  background: #EFF0F1;
+  border-radius: 2px;
+  overflow: scroll;
+  word-wrap: break-word;
+}
+
+pre::-webkit-scrollbar {
+  width: 8px;
+  height: 8px;
+  background-color: WHITE;
+}
+
+pre::-webkit-scrollbar-thumb {
+    background: grey;
+    border-radius: 10px;
+    
 }


src/main/webapp/WEB-INF/tags/base.tag
```diff
@@ -3,6 +3,6 @@
 	<h1>
 		<span class="blue">S</span><span class="red">n</span><span class="yellow">i</span><span class="blue">p</span><span class="green">p</span><span class="red">e</span><span class="yellow">t</span>
 	</h1>
-	<div class="gcse-searchbox-only" data-resultsUrl="/search"></div>
+	<div class="gcse-searchbox-only" enableAutoComplete="true" data-resultsUrl="/search"></div>
 	<a class="auth-button" href="${authUrl}">${authLabel}</a>
```
kaylaleung, line 7: Autocomplete still isn't showing up on my end :/

src/main/webapp/css/srp_style.css
```diff
@@ -22,6 +22,19 @@
   transition: box-shadow .25s;
   border-radius: 2px;
   width: 400px;
+  height: 260px;
```
kaylaleung, line 25: Change this to max-height, so cards that don't fill up 260px aren't necessarily extended.

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -21,7 +21,7 @@
             <a href="${knowledgeCard.getLink()}" title="card-source"class="card-title">
               ${knowledgeCard.getTitle()}
             </a>
-            <code>${knowledgeCard.getCode()}</code>
+            <pre><code>${knowledgeCard.getCode()}</code></pre>
             <span class="card-description">Description:</span>
```
kaylaleung, line 25: Remove "Description:" text

src/main/webapp/css/srp_style.css
```diff
@@ -43,3 +56,22 @@
   margin-top: 0;
   display: block;
 }
+
+pre {
+  background: #EFF0F1;
+  border-radius: 2px;
+  overflow: scroll;
+  word-wrap: break-word;
```
kaylaleung, line 64: I don't think this word-wrap styling changes anything based on inspect

src/main/webapp/css/srp_style.css
```diff
@@ -43,3 +56,22 @@
   margin-top: 0;
```
kaylaleung, line 56: I'd add left and right margins to .card code so the code text doesn't stick to the side of the grey area

src/main/webapp/css/srp_style.css
```diff
@@ -22,6 +22,19 @@
...
   border-radius: 2px;
   width: 400px;
+  height: 260px;
+  overflow: scroll;
+}
+
+.card::-webkit-scrollbar {
+  width: 5px;
+  height: 5px;
+  background-color: WHITE;
```
kaylaleung, line 32: lowercase?

src/main/webapp/css/srp_style.css
```diff
@@ -43,3 +56,22 @@
...
+  background: #EFF0F1;
+  border-radius: 2px;
+  overflow: scroll;
+  word-wrap: break-word;
+}
+
+pre::-webkit-scrollbar {
+  width: 8px;
+  height: 8px;
+  background-color: WHITE;
```
kaylaleung, line 70: lowercase?

src/main/webapp/css/srp_style.css
```diff
@@ -43,3 +56,22 @@
...
+}
+
+pre::-webkit-scrollbar-thumb {
+    background: grey;
+    border-radius: 10px;
+    
+}
```
kaylaleung, line 77: This is repeated code from lines 35-38

src/main/webapp/css/srp_style.css
```diff
@@ -22,6 +22,19 @@
...
   width: 400px;
+  height: 260px;
+  overflow: scroll;
+}
+
+.card::-webkit-scrollbar {
+  width: 5px;
+  height: 5px;
+  background-color: WHITE;
+}
```
kaylaleung, line 33: Add a visible scroll bar to each card to signal to the user that the card it's scrollable and not just weirdly cutoff.
```
.card::-webkit-scrollbar-thumb {
  background: lightgrey;
  border-radius: 10px;
}
```
Optionally, maybe add hover styling? At least browser scrollbars on my laptop have hover styling.
```
.card::-webkit-scrollbar-thumb:hover {
  background: grey;
  border-radius: 10px;
}
```

src/main/webapp/css/srp_style.css
```diff
@@ -22,6 +22,19 @@
   transition: box-shadow .25s;
   border-radius: 2px;
   width: 400px;
+  height: 260px;
+  overflow: scroll;
+}
+
+.card::-webkit-scrollbar {
```
kaylaleung, line 29: Just a suggestion to potentially consider but do we maybe want the hyperlinked title to stay fixed at the top of the card for easy clickability while the rest of the description/code section of the card is the scrollable only section?

I also think it seems fine this current way too though.

AaronLopes: I like the idea of a fixed title! @irisliu77 thoughts?

irisliu77: yea, i think it's better

AaronLopes: Alright, I'll open an issue for this to be done in a separate PR

src/main/webapp/WEB-INF/tags/base.tag
```diff
@@ -6,8 +6,9 @@
     </h1>
   </a>
   <div class="searchbox">
-    <div class="gcse-searchbox-only" data-resultsUrl="/search"></div>
+    <div class="gcse-searchbox-only" enableAutoComplete="true" data-resultsUrl="/search"></div>
   </div>
+	<a class="auth-button" href="${authUrl}">${authLabel}</a>
```
kaylaleung, line 11: Remove. Auth button already on line 13

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -18,10 +18,13 @@
...
-              ${knowledgeCard.getTitle()}
-            </a>
-            <pre><code>${knowledgeCard.getCode()}</code></pre>
+            <header>
+              <img class="icon" src="${knowledgeCard.getIcon()}" alt="${knowledgeCard.getSource()} site icon" width="25px" height="25px">
+              <a href="${knowledgeCard.getLink()}" title="card-source"class="card-title">
+                ${knowledgeCard.getTitle()}
+              </a>
+            </header>
+            <code>${knowledgeCard.getCode()}</code>
```
kaylaleung: You got rid of the `<pre></pre>` styling here

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -24,7 +24,7 @@
                 ${knowledgeCard.getTitle()}
               </a>
             </header>
-            <code>${knowledgeCard.getCode()}</code>
+            <pre><code>${knowledgeCard.getCode()}</code></pre>
             <span class="card-description">Description:</span>
```
irisliu77, line 28: i think this should be in <div> to avoid inline display

src/main/webapp/css/srp_style.css
```diff
@@ -42,28 +42,64 @@ h1 {
...
+}
+
+.card .card-title {
+  font: 300 20px Roboto, Arial, sans-serif;
 }
 
 .card code {
-  margin-top: 10px;
-  margin-bottom: 10px;
+  margin: 10px 5px 5px 10px;
```
kaylaleung, line 71: I think you meant for this to be `10px 5px 10px 5px` right now the 10px is being set on top and left.

src/main/webapp/css/srp_style.css
```diff
@@ -42,28 +42,64 @@ h1 {
...
+  max-height: 260px;
+  overflow: scroll;
+  font: 16px Roboto, Arial, sans-serif;
+  margin-top: 20px;
 }
 
-.card .card-title {
-  font: 300 20px Roboto, Arial, sans-serif;
+.card::-webkit-scrollbar {
+  width: 8px;
```
kaylaleung, line 52: nit: Since the scrollbar is 8px, I think it looks better if instead of .card padding being 24px all around, it's slightly adjusted for the right hand side with `24px 16px 24px 24px` but this is slightly eyeballing and maybe not completely precise

src/main/webapp/WEB-INF/tags/base.tag
```diff
@@ -6,7 +6,7 @@
     </h1>
   </a>
   <div class="searchbox">
-    <div class="gcse-searchbox-only" data-resultsUrl="/search"></div>
+    <div class="gcse-searchbox-only" enableAutoComplete="true" data-resultsUrl="/search"></div>
```
joanchenx, line 9: I know @kaylaleung mentioned she wasn't able to get the suggestions to show up for her demo. I don't seem to be able to either.

Kayla - were you able to fix it and see suggestions?
Aaron - are you able to see suggestions?

src/main/webapp/css/srp_style.css
```diff
@@ -37,33 +37,69 @@ h1 {
...
+pre::-webkit-scrollbar {
+  width: 5px;
+  height: 5px;
+  background: #FAFAFA;  
+}
+
+pre::-webkit-scrollbar-thumb {
+    background: lightgrey;
+    border-radius: 10px;
+
```
joanchenx, line 96: Remove new line.

src/main/webapp/WEB-INF/tags/base.tag
AaronLopes: Still no suggestions yet :/

joanchenx: Got it, could you edit the PR description so that it doesn't say that it fixes #75?

Also, when I looked into this, it seems like this is something you also need to enable from the control panel?
https://blog.expertrec.com/how-to-enable-autocomplete-in-google-custom-search/

But in my control panel, I only see the three search engines for each of the clients and not the actual project CSE. So I unfortunately don't know how to get this to work :'( Not sure if I am missing something/if you're able to figure out the control panel?

AaronLopes: Autocomplete was enabled (by default) in the control panel since creating the key; adding this attribute was an extra step :')

joanchenx: got it, merp yeah i don't have a clue on why it's not working

src/main/webapp/WEB-INF/templates/search.jsp
```diff
@@ -24,10 +24,10 @@
                 ${knowledgeCard.getTitle()}
               </a>
             </header>
+            <pre><code>${knowledgeCard.getCode()}</code></pre>
```
anbenson, line 27: Code is appearing twice now (bad merge?).

Do you mean
```html
<c:if test="${knowledgeCard.getCode() != null}">
  <pre><code>${knowledgeCard.getCode()}</code></pre>
</c:if>
```

src/main/webapp/WEB-INF/tags/base.tag
anbenson, line 9: Let's be really careful that this PR does not automatically close the searchbox suggestions issue. I'm not sure if it will or not, but the phrase "fix 75" (with the hashtag) does appear in the PR (albeit prefixed with "did not"), so it might get closed anyhow.

src/main/webapp/css/srp_style.css
```diff
@@ -37,33 +37,68 @@ h1 {
...
   width: 400px;
+  max-height: 260px;
+  overflow: scroll;
+  font: 16px Roboto, Arial, sans-serif;
+  margin-top: 20px;
 }
 
-.card .card-title {
-  font: 300 20px Roboto, Arial, sans-serif;
+.card::-webkit-scrollbar {
```
anbenson, line 51: This is non-standard CSS. The standards-compliant (ish) one is https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars, but it's only supported by Firefox. Can you use both for coverage?

src/main/webapp/css/srp_style.css
```diff
@@ -37,33 +37,68 @@ h1 {
 
 .card {
   margin-bottom: 20px;
-  padding: 24px;
+  padding: 24px 16px 24px 24px;
```
anbenson, line 40: Why? This looks sketchy.

src/main/webapp/css/srp_style.css
```diff
@@ -37,33 +37,68 @@ h1 {
...
   box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 3px 1px -2px rgba(0, 0, 0, 0.12), 0 1px 5px 0 rgba(0, 0, 0, 0.2);
   transition: box-shadow .25s;
   border-radius: 2px;
   width: 400px;
+  max-height: 260px;
+  overflow: scroll;
+  font: 16px Roboto, Arial, sans-serif;
+  margin-top: 20px;
```
anbenson, line 48: You already have `margin-bottom`. Maybe replace with `margin: 20px 0`? (Put whatever margin property(s) with padding so similar properties are near each other.)

src/main/webapp/css/srp_style.css
```diff
@@ -37,33 +37,68 @@ h1 {
...
   margin-bottom: 20px;
-  padding: 24px;
+  padding: 24px 16px 24px 24px;
   box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 3px 1px -2px rgba(0, 0, 0, 0.12), 0 1px 5px 0 rgba(0, 0, 0, 0.2);
   transition: box-shadow .25s;
   border-radius: 2px;
   width: 400px;
+  max-height: 260px;
+  overflow: scroll;
+  font: 16px Roboto, Arial, sans-serif;
```
anbenson, line 47: This is already inherited; this does nothing.

src/main/webapp/css/srp_style.css
```diff
@@ -37,33 +37,68 @@ h1 {
...
+}
+
+.card .card-title {
+  font: 300 20px Roboto, Arial, sans-serif;
 }
 
 .card code {
-  margin-top: 10px;
-  margin-bottom: 10px;
+  margin: 10px 5px 10px 5px;
```
anbenson, line 71: This is equivalent to

`margin: 10px 5px`

src/main/webapp/css/srp_style.css
```diff
@@ -37,33 +37,68 @@ h1 {
...
+
+pre::-webkit-scrollbar-thumb {
+  background: lightgrey;
+  border-radius: 10px;    
+}
+
+pre::-webkit-scrollbar-thumb:hover {
+  background: grey; 
+}
+
```
anbenson, line 101: This code seems familiar...can you combine it? Either in one rule, or make this a separate class.

src/main/webapp/css/srp_style.css
kaylaleung: The scroll bar on the right takes up 8 px of the container? But a pretty minor UI thing I guess.

anbenson: I should point out that this is particular to Chrome. (See the non-standard "overflow: overlay" property to see how you could make Chrome and other Chrome-like browsers not make the scrollbar take up additional space, and also "scrollbar-gutter" for the \~future~ of controlling this.) Other browsers may not have this - for example, Firefox on Mac puts the scrollbar on top of the padding, so it looks off-center there.

src/main/webapp/css/srp_style.css
```diff
@@ -36,34 +36,61 @@ h1 {
...
 
-.card .card-description {
-  font: bold 16px Roboto, Arial, sans-serif;
-  margin-top: 20px;
+.card::-webkit-scrollbar-thumb:hover, pre::-webkit-scrollbar-thumb:hover {
+  background: grey;
+}
+
+.card .card-title {
+  font: 300 20px Roboto, Arial, sans-serif;
```
anbenson, line 67: Same here, font is already inherited.

src/main/webapp/css/srp_style.css
```diff
@@ -36,34 +36,61 @@ h1 {
...
+  background: #EFF0F1;
+  border-radius: 2px;
+  overflow: scroll;
+  scrollbar-width: thin;
+}
+
+pre::-webkit-scrollbar {
+  width: 5px;
+  height: 5px;
+  background: #FAFAFA;
```
anbenson, line 91: scrollbar-color for pre?

src/main/webapp/css/srp_style.css
AaronLopes: is there a way to combine font size and weight without including the font family at the end?

anbenson: Nope.

https://developer.mozilla.org/en-US/docs/Web/CSS/font
If font is specified as a shorthand for several font-related properties, then:

    it must include values for:
        <font-size>
        <font-family>

PR #82, opened 2020-07-25
Account page setting UI

irisliu77:
create a page for setting up preferred websites and coding languages. navigate to the page through /user


.eslintrc.json
```diff
@@ -9,6 +9,7 @@
   },
   "rules": {
     "require-jsdoc": "off",
-    "valid-jsdoc": "off"
+    "valid-jsdoc": "off",
+    "no-unused-vars": "off"
```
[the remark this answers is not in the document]
irisliu77: this is not an gitignore file, this needs to be changed in order to pass the eslint check on github

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset="utf-8">
+    <title>User Dashboard</title>
+    <script src="../../data/languages.js"></script>
+    <script src="../../js/user.js"></script>
```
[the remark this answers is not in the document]
irisliu77: this does not work, but i change to `data/languages.js` and  `js/user.js` and it works.

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,38 @@
...
+                <option value="geeksforgeeks">GeeksForGeeks</option>
+                <option value="stackoverflow">StackOverflow</option>
+                <option value="w3schools">W3Schools</option>
+            </select>
+        </div>
+        <div id="languages" class="column">
+            <label for="input-languages">Choose your primary coding language:</label>
+            <input id="input-languages" type="text" name="input-languages"
+                   placeholder="Input your primary coding language">
+            <script>autocomplete(document.getElementById('input-languages'), languages);</script>
```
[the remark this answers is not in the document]
irisliu77: i cannot import a js file (languages.js) to another js file (user.js) based on current setting of the package.json so i cannot move this line into the js file. let me know if you figure out any way to do it.

src/main/webapp/css/user_style.css
```diff
@@ -0,0 +1,89 @@
...
+}
+
+#setting-submit {
+    display: block;
+    margin-top: 5%;
+    margin-left: 47.5%;
+}
+
+#websites {
+    vertical-align: top;
```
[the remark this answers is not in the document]
irisliu77: the autosuggest of languages will stretch the div and make `#websites` falls to the bottom of the div. this vertical-align makes it stick to the top of the div

src/main/webapp/js/user.js
```diff
@@ -0,0 +1,75 @@
+function autocomplete(inp, arr) {
+  let currentFocus;
+  inp.addEventListener('input', function(e) {
+    let a = this.value;
+    let b = this.value;
+    const val = this.value;
```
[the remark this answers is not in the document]
irisliu77: ui bug is solved and i changed some variable names

src/main/webapp/data/languages.js
```diff
@@ -0,0 +1,707 @@
+const languages = [
+  'A# .NET',
```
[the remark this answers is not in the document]
anbenson: Handpick 10 or so maybe? Then it doesn't even need to be a separate file.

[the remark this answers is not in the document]
irisliu77: decided to keep this long list after discussion

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,38 @@
...
+        </div>
+        <input id="setting-submit" type="submit" value="Submit">
+    </form>
+</div>
```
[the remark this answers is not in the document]
kaylaleung: unresolved?

src/main/webapp/css/user_style.css
```diff
@@ -0,0 +1,84 @@
...
+.autocomplete-items {
+    border: 1px solid #d4d4d4;
+    border-bottom: none;
+    border-top: none;
+    z-index: 99;
+    top: 100%;
+    left: 0;
+    right: 0;
+    max-height: 300px;
+    overflow: scroll;
```
kaylaleung, line 43: There's still some weird styling happening when the scroll bar disappears and there's just an awkward bottom and right border around the autocomplete items dropdown

I think the bottom border can be resolved by making this overflow-y: scroll. 

You might want to look into some of the `autocomplete-items::-webkit-scrollbar` styling, but I guess I'm just nitpicking at UI style at this point

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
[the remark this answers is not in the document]
irisliu77: wait, i think i changed it and my formatter in ide format it again

src/main/webapp/css/user_style.css
```diff
@@ -0,0 +1,83 @@
...
+    color: #fff;
+}
+
+.autocomplete-items {
+    border: 1px solid #d4d4d4;
+    border-top: none;
+    z-index: 99;
+    top: 100%;
+    left: 0;
+    right: 0;
```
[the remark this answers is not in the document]
irisliu77: yea, i removed it

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+<head>
```
anbenson, line 3: I think it'd be nice to have the Snippet logo on this page as well? It feels pretty empty.

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,32 @@
...
+    <meta charset="utf-8">
+    <title>User Dashboard</title>
+    <script src="data/languages.js"></script>
+    <script src="js/user.js"></script>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+    <link rel="stylesheet" type="text/css" href="/css/user_style.css">
```
anbenson, line 9: type="text/css" isn't necessary

anbenson, line 9: href="css/user_style.css"

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,32 @@
...
+</head>
+<body>
+  <div class="main">
+      <form class="setting" action="/user" method="POST">
```
anbenson, line 13: Indentation is inconsistent. Line 12 is 2 spaces. Line 13 is 4 spaces. I think we've generally done 2.

src/main/webapp/css/user_style.css
```diff
@@ -0,0 +1,83 @@
+.main {
+    text-align: center;
```
anbenson, line 2: The other CSS files are 2 space tabbed

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>User Dashboard</title>
+    <script src="data/languages.js"></script>
+    <script src="js/user.js"></script>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
```
anbenson, line 8: I don't think jQuery is used?

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,32 @@
...
+                  <option value="geeksforgeeks">GeeksForGeeks</option>
+                  <option value="stackoverflow">StackOverflow</option>
+                  <option value="w3schools">W3Schools</option>
+              </select>
+          </div>
+          <div id="languages" class="column">
+              <label for="input-languages">Choose your primary coding language:</label>
+              <input class="input-languages" id="input-languages" type="text" name="input-languages"
+                     placeholder="Input your primary coding language">
+              <script>autocomplete(document.getElementById('input-languages'), languages);</script>
```
anbenson, line 26: Instead of inline scripts, can you just make the code be run on DOMContentLoaded? (See how https://github.com/googleinterns/step1-2020/pull/76/files does it)

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,32 @@
...
+<head>
+    <meta charset="utf-8">
+    <title>User Dashboard</title>
+    <script src="data/languages.js"></script>
+    <script src="js/user.js"></script>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+    <link rel="stylesheet" type="text/css" href="/css/user_style.css">
+</head>
+<body>
+  <div class="main">
```
anbenson, line 12: This div doesn't really do anything that body doesn't. If you need to add font-family CSS or whatnot to make it global, add it to body.

src/main/webapp/css/user_style.css
```diff
@@ -0,0 +1,83 @@
+.main {
```
anbenson, line 1: Can you sort the CSS in this file in order of appearance in the HTML? It's taking me quite a while to find the rules I'm looking for.

src/main/webapp/css/user_style.css
```diff
@@ -0,0 +1,83 @@
...
+    background-color: #e9e9e9;
+}
+
+.autocomplete-active {
+    background-color: #4285F4;
+    color: #ffffff;
+}
+
+.setting {
+    margin-top: 20%;
```
anbenson, line 61: I don't think this does what you think it does. Percentage-based margins are based on the container's width! Are you trying to make this responsive with height? Or would a pixel-based margin work better?

src/main/webapp/css/user_style.css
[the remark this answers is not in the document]
anbenson: This doesn't seem like the right solution. Autocomplete suggestions are generally a layer that hovers over other content. I would aim to make the HTML for your suggestions position: absolute.

src/main/webapp/css/user_style.css
```diff
@@ -0,0 +1,83 @@
...
+    border: 1px solid transparent;
+    background-color: #f1f1f1;
+    padding: 10px;
+    font-size: 16px;
+    font-family: Roboto, Arial, sans-serif;
+}
+
+.setting-submit {
+    display: block;
+    margin-top: 5%;
```
anbenson, line 77: Again, this probably doesn't do what you think it does.

src/main/webapp/css/user_style.css
```diff
@@ -0,0 +1,83 @@
...
+    margin-left: 47.5%;
```
anbenson, line 78: Can you just center it instead of using percentage-based margins?

anbenson: In general the layout of this form should be revisited. Do you really want the select/input fields to be half the width of the screen? That's going to be way too large in general on most desktop/laptop devices. How about setting a max-width on these inputs, and designing your form's layout to have two centered rows?

src/main/webapp/js/user.js
```diff
@@ -0,0 +1,91 @@
...
+  input.addEventListener('input', function(e) {
+    let autocomplete = this.value;
+    let language = this.value;
+    const val = this.value;
+    closeAllLists();
+    if (!val) {
+      return false;
+    }
+    currentFocus = -1;
+    autocomplete = document.createElement('div');
```
anbenson, line 12: Instead of generating HTML in JS, consider adding a display: none HTML element in the HTML that represents what a suggestion looks like (including all the ids and classes you need), then using https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode to copy it. It'll be easier to maintain and visualize.

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,32 @@
...
+              <label for="input-websites">Choose your preferred website:</label>
+              <select class="input-websites" id="input-websites" name="input-websites">
+                  <option value="geeksforgeeks">GeeksForGeeks</option>
+                  <option value="stackoverflow">StackOverflow</option>
+                  <option value="w3schools">W3Schools</option>
+              </select>
+          </div>
+          <div id="languages" class="column">
+              <label for="input-languages">Choose your primary coding language:</label>
+              <input class="input-languages" id="input-languages" type="text" name="input-languages"
```
anbenson, line 24: Suggestions don't take up the full width of the input.

src/main/webapp/js/user.js
```diff
@@ -0,0 +1,91 @@
...
+        language.addEventListener('click', function(e) {
+          input.value = this.getElementsByTagName('input')[0].value;
+          closeAllLists();
+        });
+        autocomplete.appendChild(language);
+      }
+    }
+  });
+  /* Execute a function presses a key on the keyboard */
+  input.addEventListener('keydown', function(e) {
```
anbenson, line 34: Implementing keyboard accessibility is a nice touch, but it's hard to do. There's a bunch of bugs I see here. For one, you can press the down arrow past the visible options.

Also note that for actual accessibility, you should be changing what element has focus as you press keys. Then you don't need to handle Enter (and Space, which was omitted, will work just like that). This also makes is so that the event `e` tells you what element was previously focused, so you don't have to deal with all this autocomplete-active maintenance.

src/main/webapp/js/user.js
```diff
@@ -0,0 +1,91 @@
...
+    const items = document.getElementsByClassName('autocomplete-items');
+    for (let i = 0; i < items.length; i++) {
+      if (elmnt != items[i] && elmnt != input) {
+        items[i].parentNode.removeChild(items[i]);
+      }
+    }
+  }
+
+  /* Execute a function when someone clicks in the document */
+  document.addEventListener('click', function(e) {
```
anbenson, line 88: Global click listeners are not scalable. Instead, keep element focus in the list when it's open, and check for the blur event.

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
irisliu77: i've done this before, but seems like i cannot import the languages.js file to user.js based on current project setting.

src/main/webapp/css/user_style.css
```diff
@@ -0,0 +1,39 @@
...
+  border: 1px solid transparent;
+  background-color: #f1f1f1;
+  padding: 10px;
+  font-size: 16px;
+  font-family: Roboto, Arial, sans-serif;
+}
+
+.setting-submit {
+  margin: auto;
+}
```
joanchenx, line 39: I don't think this is used anywhere?

src/main/webapp/css/user_style.css
```diff
@@ -0,0 +1,39 @@
...
+  margin: 25px auto;
+}
+
+.input-websites, input[type=text] {
+  width: 100%;
+  border: 1px solid transparent;
+  background-color: #f1f1f1;
+  padding: 10px;
+  font-size: 16px;
+  font-family: Roboto, Arial, sans-serif;
```
joanchenx, line 34: Combine the font setting into a single line: `font: 16px Roboto, Arial, sans-serif`

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,39 @@
...
+        <span class="blue">S</span><span class="red">n</span><span class="yellow">i</span><span
+          class="blue">p</span><span class="green">p</span><span class="red">e</span><span
+          class="yellow">t</span>
+      </h1>
+    </a>
+  </div>
+  <meta charset="utf-8">
+  <title>User Dashboard</title>
+  <link rel="stylesheet" href="css/style.css">
+  <link rel="stylesheet" href="css/user_style.css">
```
joanchenx, line 19: Move these to right under `<head>`, line 7

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,39 @@
...
+        <select class="input-websites" id="input-websites" name="website">
+          <option value="geeksforgeeks">GeeksForGeeks</option>
+          <option value="stackoverflow">StackOverflow</option>
+          <option value="w3schools">W3Schools</option>
+        </select>
+      </div>
+      <div class="row">
+        <label for="input-languages">Choose your primary coding language:</label>
+        <input class="input-languages" id="input-languages" type="text" name="language"
+               placeholder="Input your primary coding language">
```
joanchenx, line 34: It seems like the bottom language text input width is larger than the website select. Could you take a look?

src/main/java/com/google/step/snippet/servlets/UserSettingServlet.java
```diff
@@ -0,0 +1,34 @@
...
+package com.google.step.snippet.servlets;
+
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servelet that renders user setting page */
```
anbenson, line 24: Servlet

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
irisliu77: i'm not sure whether i used the correct approach to solve this issue..

irisliu77: i added `width: -webkit-fill-available` to both input field

anbenson: No, anything prefixed with -webkit- is a hack that only works in Chrome and similar Chrome-like browsers. Avoid those except when absolutely necessary.

The crux of your issue is that &lt;select&gt; and &lt;input&gt; behave differently in their `box-sizing` property. &lt;select&gt; is `border-box`, which means its width includes padding and border, whereas &lt;input&gt; is `content-box`, which means width does not include padding and border.

Since your div gets a max-width of 400px, the select element gives its content only 378px so that the remaining 22px is allocated to its 10px padding and 1px border (on each side). The input element instead gives 400px to its content, but the extra 22px needed by the padding and border given causes it to go past the 400px max you've given its container div.

To fix, simply have `.input-languages` have `box-sizing: border-box;`.

src/main/webapp/WEB-INF/templates/user_dashboard.jsp
```diff
@@ -0,0 +1,39 @@
+<%@page contentType="text/html" pageEncoding="UTF-8" %>
+<%@taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
+
+<!DOCTYPE html>
+<html>
+<head>
```
anbenson, line 6: &lt;head&gt; and &lt;body&gt; should be indented within &lt;html&gt;

src/main/webapp/css/user_style.css
```diff
@@ -0,0 +1,38 @@
...
+}
+
+.row {
+  max-width: 400px;
+  margin: 25px auto;
+}
+
+.input-websites, input[type=text] {
+  width: 100%;
+  border: 1px solid transparent;
```
anbenson, line 30: Having a transparent border seems really weird. Since this is under border-box, it becomes effectively part of the padding.

It looks like all you're trying to do is get rid of the ugly border. In that case, do `border: none`.

Issue #80, opened 2020-07-24
Uninformative w3schools description


kaylaleung:
W3Schools scrapper for sites such as [https://www.w3schools.com/css/css3_flexbox.asp](https://www.w3schools.com/css/css3_flexbox.asp) returns unhelpful description such as "Try It Yourself".

Should to a string compare chec if it's one of these unhelpful generic descriptions and if so take the next paragraph description instead of .first

PR #83, opened 2020-07-27
Improve description retrieval for W3Schools

kaylaleung:
Increase description retrieval specificity to avoid returning "Try It Yourself" as a card description. Addresses #80

### DIFF
--- a/src/main/java/com/google/step/snippet/external/W3SchoolsClient.java
+++ b/src/main/java/com/google/step/snippet/external/W3SchoolsClient.java
@@ -56,10 +56,12 @@ public final class W3SchoolsClient implements Client {
     if (snippets.isEmpty() || snippets.first().getElementsByClass(CODE_CLASS).text().isEmpty()) {
       return null;
     }
-    String title = titles.first().text();
-    String description = descriptions.first().text();
+    String title = Jsoup.clean(titles.first().text(), Whitelist.relaxed());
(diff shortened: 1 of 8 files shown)


irisliu77:
so the only change here in this pr is in the W3SchoolsClient? (or maybe also its test?)

kaylaleung:
Just the client, depending on if @AaronLopes includes the escaping tests in instead #74

anbenson:
Can you rebase this on top of master so it's more clear what new changes this PR contains?

kaylaleung:
> Can you rebase this on top of master so it's more clear what new changes this PR contains?

Waiting for #74  to be merged

src/main/java/com/google/step/snippet/external/W3SchoolsClient.java
```diff
@@ -57,7 +57,17 @@ public Card search(String w3Link, String query) {
       return null;
     }
     String title = Jsoup.clean(titles.first().text(), Whitelist.relaxed());
-    String description = Jsoup.clean(descriptions.first().text(), Whitelist.relaxed());
+    String description;
+    /* If description for specific code example exists, use example description, otherwise use first
+    page description */
```
joanchenx, line 62: Nit: I would suggest using double-slash comments for comments inside your methods and add a period at the end of the sentence.

```
// If description for specific code example exists, use example description, otherwise use first
// page description.
```

Issue #85, opened 2020-07-27
Add tests for HTML escaping


AaronLopes:
Tests need to be created for clients which take into account HTML escaping on certain queries.

PR #87, opened 2020-07-28
Add HTML escaping tests

kaylaleung:
Address #85

### DIFF
--- a/src/test/java/com/google/step/snippet/external/W3SchoolsClientTest.java
+++ b/src/test/java/com/google/step/snippet/external/W3SchoolsClientTest.java
@@ -44,6 +44,36 @@ public final class W3SchoolsClientTest {
   }
 
   @Test
+  public void cssEscaping() {
+    Card actual = client.search("https://www.w3schools.com/html/html_css.asp", "css style");
+    Card expected =
+        new Card(
+            "HTML Styles - CSS",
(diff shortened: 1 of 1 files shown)


src/test/java/com/google/step/snippet/external/W3SchoolsClientTest.java
```diff
@@ -43,6 +43,36 @@ public void jsonCodeCard() {
     assertEquals(expected, actual);
   }
 
+  @Test
+  public void cssEscaping() {
+    Card actual = client.search("https://www.w3schools.com/html/html_css.asp", "css style");
```
joanchenx, line 48: I think the logic for determining when to do HTML escaping should also be checked. Could you add tests where you check:
1. Query does not contain escape filter, but webpage does. In this case, the HTML should still be escaped. 
Example: `client.search("https://www.w3schools.com/html/html_css.asp", "style");`
2. Query contains escape filter, but webpage does not. In this case, the HTML should still be escaped.
Example: `client.search("https://www.w3schools.com/js/js_errors.asp", "html js error");`
3. Neither query nor webpage contains escape filter. In this case, the HTML should not be escaped.
Example: `client.search("https://www.w3schools.com/js/js_errors.asp", "js error");`

PR #92, opened 2020-07-28
Implement sorted display

AaronLopes:
get user site preferences and respectively sort cards


src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -89,11 +98,12 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     if (request.getQueryString() != null) {
       redirectPath += "?" + request.getQueryString();
     }
-
+    String pSite = "";
```
irisliu77, line 101: `pSite` is a bit unclear

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -136,7 +146,18 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     } catch (InterruptedException | RejectedExecutionException e) {
       allCards = Collections.emptyList();
     }
-
+    if (!preferredSite.isEmpty()) {
+      List<Card> sortedCards = allCards;
+      for (Card card : sortedCards) {
+        if (card.getSource().toLowerCase().equals(preferredSite)) {
+          int index = allCards.indexOf(card);
+          sortedCards.remove(index);
```
irisliu77, line 154: add `if (index != 0)` here, if the index already equals to 0, no further action is needed

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -136,7 +146,18 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     } catch (InterruptedException | RejectedExecutionException e) {
       allCards = Collections.emptyList();
     }
-
+    if (!preferredSite.isEmpty()) {
+      List<Card> sortedCards = allCards;
```
irisliu77, line 150: i don't think a new list of card is needed here, you can just do those operation on `allCards`

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -89,11 +98,12 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     if (request.getQueryString() != null) {
       redirectPath += "?" + request.getQueryString();
     }
-
+    String preferredSite = "";
     UserService userService = UserServiceFactory.getUserService();
     if (userService.isUserLoggedIn()) {
       request.setAttribute(AUTH_URL, userService.createLogoutURL(redirectPath));
       request.setAttribute(AUTH_LABEL, "Logout");
+      preferredSite = getPreferredSite(userService.getCurrentUser().getUserId());
```
irisliu77, line 106: preferredSite needs `toLowerCase()` here as respect to line 152

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
AaronLopes: Doing that could result in a ConcurrentAccessException, which is why I copied the list.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
AaronLopes: getPreferredSite already returns the source in lowercase

irisliu77: uh, right, i forgot that i've already set all the value in the select div to be lower case

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -176,4 +199,17 @@ private String getLink(String id, String query) {
...
+
+  private String getPreferredSite(String userId) {
+    Query.FilterPredicate filterId =
+        new Query.FilterPredicate(ID_PARAMETER, FilterOperator.EQUAL, userId);
+    Query userQuery = new Query(USER_PARAMETER).setFilter(filterId);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Entity userPreferenceEntity = datastore.prepare(userQuery).asSingleEntity();
+    String site = "";
+    if (userPreferenceEntity != null) {
+      site = (String) userPreferenceEntity.getProperty(WEBSITE_PARAMETER);
```
kaylaleung, line 211: Add null check here in case there isn't a website preference set.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -136,7 +146,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     } catch (InterruptedException | RejectedExecutionException e) {
       allCards = Collections.emptyList();
     }
-
+    if (!preferredSite.isEmpty()) {
```
anbenson, line 149: Can we put a newline before this line for some visual separation between these steps?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -136,7 +146,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+      List<Card> sortedCards = allCards;
```
anbenson, line 150: You're mutating the list, so I don't see any reason to do aliasing.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -136,7 +146,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+      for (Card card : sortedCards) {
+        if (card.getSource().toLowerCase().equals(preferredSite)) {
+          int index = allCards.indexOf(card);
```
anbenson, line 153: Sounds like you need the index. How about doing a normal for loop (with indices) - that way you don't need to get the index using this O(n) operation?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -137,6 +148,21 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
       allCards = Collections.emptyList();
     }
 
+    if (!preferredSite.isEmpty()) {
+      List<Card> sortedCards = new ArrayList<>(allCards);
```
joanchenx, line 152: Do you need `sortedCards`? Can you just use `allCards`?

I saw in another comment you mentioned that doing so could result in a ConcurrentAccessException. CMIIW, but I think by not using the for-each loop, you won't run into this issue anymore?

PR #95, opened 2020-07-28
Add primary language feature

irisliu77:


src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -176,4 +204,22 @@ private String getLink(String id, String query) {
...
+  private String[] getPreference(String userId) {
+    String[] result = new String[2];
+    Query.FilterPredicate filterId =
+        new Query.FilterPredicate(ID_PARAMETER, FilterOperator.EQUAL, userId);
+    Query userQuery = new Query(USER_PARAMETER).setFilter(filterId);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Entity userPreferenceEntity = datastore.prepare(userQuery).asSingleEntity();
+    if (userPreferenceEntity != null) {
+      result[0] = (String) userPreferenceEntity.getProperty(WEBSITE_PARAMETER);
+      result[1] = (String) userPreferenceEntity.getProperty(LANGUAGE_PARAMETER);
```
kaylaleung, line 217: check properties are non null?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -200,17 +207,23 @@ private String getLink(String id, String query) {
     return null;
   }
 
-  private String getPreferredSite(String userId) {
+  private String[] getPreference(String userId) {
+    String[] result = new String[2];
```
anbenson, line 211: This changes the API. Before, it would return "" when a preference was not found, but now it returns null. This would cause a NPE in Aaron's code, which assumes it would only be "", not null. Can you initialize to a literal array of two empty strings?

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -200,17 +207,23 @@ private String getLink(String id, String query) {
...
         new Query.FilterPredicate(ID_PARAMETER, FilterOperator.EQUAL, userId);
     Query userQuery = new Query(USER_PARAMETER).setFilter(filterId);
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     Entity userPreferenceEntity = datastore.prepare(userQuery).asSingleEntity();
-    String site = "";
     if (userPreferenceEntity != null
-        && userPreferenceEntity.getProperty(WEBSITE_PARAMETER) != null) {
-      site = (String) userPreferenceEntity.getProperty(WEBSITE_PARAMETER);
+        && userPreferenceEntity.getProperty(WEBSITE_PARAMETER) != null
+        && userPreferenceEntity.getProperty(LANGUAGE_PARAMETER) != null) {
```
anbenson, line 219: It's possible for one but not the other to be null. Can you make these checks separate, so that you only set result[i] if it's not null?

PR #96, opened 2020-07-29
GeeksForGeeks Tests

AaronLopes:
(Ignore few SO tests for now, too many API calls today)

### DIFF
--- a/src/test/java/com/google/step/snippet/external/GeeksForGeeksClientTest.java
+++ b/src/test/java/com/google/step/snippet/external/GeeksForGeeksClientTest.java
@@ -87,4 +87,97 @@ public final class GeeksForGeeksClientTest {
     Card actual = client.search("https://www.geeksforgeeks.org/", "geeksforgeeks");
     assertNull(actual);
   }
+
+  @Test
+  public void escapeHtmlPage() {
+    Card actual =
+        client.search(
+            "https://www.geeksforgeeks.org/how-to-change-the-height-of-br-tag/", "br tag");
+    Card expected =
+        new Card(
+            "How to change the height of br tag?",
+            "&lt;!DOCTYPE html &gt; &lt;html&gt; &lt;head&gt; &lt;title&gt; Customized break"
+                + " example &lt;/title&gt; &lt;style type=&quot;text/css&quot;&gt; .br { display:"
+                + " block; margin-bottom: 0em; } .brmedium { display: block; margin-bottom: 1em; }"
+                + " .brlarge { display: block; margin-bottom: 2em; } &lt;/style&gt; &lt;/head&gt;"
+                + " &lt;body&gt; &lt;h3&gt;This page shows different break height between"
+                + " lines&lt;/h3&gt; &lt;p&gt; Hi User &lt;span"
+                + " class=&quot;brlarge&quot;&gt;&lt;/span&gt; Welcome to &lt;span"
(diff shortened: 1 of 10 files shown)


anbenson:
Remind me who's already approved this? Is it just me who's left, or should we be waiting for the pod to approve and so forth?

irisliu77:
> Remind me who's already approved this? Is it just me who's left, or should we be waiting for the pod to approve and so forth?

the pod approved it, i think Joan also needs to be added here

src/test/java/com/google/step/snippet/external/StackOverflowClientTest.java
```diff
@@ -100,7 +100,7 @@ public void noAnswerSearchTest() {
     Card actual =
         client.search(
             "https://stackoverflow.com/questions/44686609/implementing-a-neural-network-in-haskell",
-            "implement neural net in haskell");
+            "implement neural network in haskell");
```
anbenson, line 103: Was this an intended change?

src/test/java/com/google/step/snippet/external/GeeksForGeeksClientTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class GeeksForGeeksClientTest {
+
+  private final GeeksForGeeksClient client = new GeeksForGeeksClient("CSE_ID");
+
+  @Test
+  public void htmlCodeCard() {
```
joanchenx, line 17: I think the logic for determining when to do HTML escaping should also be checked. Could you add tests where you check:
1. Query does not contain escape filter, but webpage does. In this case, the HTML should still be escaped. 
Example: `client.search("https://www.geeksforgeeks.org/how-to-change-the-height-of-br-tag/", "br tag");`
2. Query contains escape filter, but webpage does not.  In this case, the HTML should still be escaped. 
Example: `client.search("https://www.geeksforgeeks.org/how-to-display-error-without-alert-box-using-javascript/", "html display error");`
3. Neither query nor webpage contains escape filter. In this case, the HTML should not be escaped. 
Example: `client.search("https://www.geeksforgeeks.org/how-to-display-error-without-alert-box-using-javascript/", "javascript display error");`

AaronLopes: Sure!

src/test/java/com/google/step/snippet/external/StackOverflowClientTest.java
AaronLopes: No, I think it was from a previous edit I made to the file. Removed.

src/test/java/com/google/step/snippet/external/GeeksForGeeksClientTest.java
```diff
@@ -0,0 +1,183 @@
...
+                + " </div>  Submit  \n"
+                + "</div>",
+            "https://www.geeksforgeeks.org/how-to-display-error-without-alert-box-using-javascript/",
+            "Errors in JavaScript can be displayed without the use of alert boxes but using the"
+                + " alert box is the traditional way to do that. We can show errors with two"
+                + " methods without using the alert box.",
+            "GeeksForGeeks",
+            "https://geeksforgeeks.org/favicon.ico");
+    assertEquals(expected, actual);
+  }
```
joanchenx, line 182: Great! Small nit: could you move these underneath the `htmlCodeCard` test?

PR #99, opened 2020-07-29
Implement Voting Feedback with Authentication

### LINKED ISSUE #86: Thread safety for Datastore
Currently, Cards aren't being loaded on branches which make use of Datastore functions (eg. Feedback). Issue will be updated with resolution progress.

Error: `java.lang.NullPointerException: No API environment is registered for this thread.`

kaylaleung:
Fix #86


src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,177 @@
...
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    // Invalid POST request if URL is invalid or user is not logged
+    String url = request.getParameter(URL);
+    UserService userService = UserServiceFactory.getUserService();
+    if (url == null || !userService.isUserLoggedIn()) {
+      JSONObject json = new JSONObject();
+      response.setContentType("application/json");
+      response.getWriter().println(json);
```
irisliu77, line 84: i think adding a notice asking the user to log in will be better than returning an empty json?

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,177 @@
...
+    } else if (request.getParameter(DOWN) != null) {
+      if (!downCards.contains(url)) {
+        downCards.add(url);
+        if (upCards.contains(url)) {
+          upCards.remove(url);
+        }
+      } else {
+        downCards.remove(url);
+      }
+    }
```
irisliu77, line 124: for the logic here, i want to know when a user click downvote when he already upvoted the exact same card, will the status become 0 or -1?

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
kaylaleung: I can add a message popup on the javascript end when it finds the empty json to ask the user to login

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
kaylaleung: If the user previously upvoted and then toggles downvote, it'll become -1 because the upvote is removed and the downvote is added.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
```diff
@@ -122,7 +126,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
     List<Card> allCards;
     try {
       allCards =
-          executor.invokeAll(cardCallbacks, 3L, TimeUnit.SECONDS).stream()
+          executor.invokeAll(cardCallbacks, 30L, TimeUnit.SECONDS).stream()
```
AaronLopes, line 129: how was this timeout decided?

kaylaleung: Uh not all the cards were coming up all the time after adding the query to datastore so the timing was bumped up.

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,174 @@
...
+      downCards = new ArrayList<String>();
+    } else {
+      downCards = (ArrayList<String>) userEntity.getProperty(DOWNVOTED);
+    }
+
+    // Add or remove card from user's list of upvoted and downvoted cards
+    if (request.getParameter(UP) != null) {
+      if (!upCards.contains(url)) {
+        upCards.add(url);
+        if (downCards.contains(url)) {
```
anbenson, line 108: This guard isn't necessary, as remove is a no-op if it's not in the list.

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,174 @@
...
+          downCards.remove(url);
```
anbenson, line 109: This doesn't seem necessary, as the lack of DOWN as a request parameter already removes it.

In other words:
If UP exists, add it, else remove it.
If DOWN exists, add it, else remove it.

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,174 @@
...
+      response.setContentType("application/json");
+      response.getWriter().println(json);
+      return;
+    }
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Entity userEntity = getUserEntity(datastore, userService);
+
+    // Get lists of user's upvoted and downvoted cards
+    ArrayList<String> upCards;
```
anbenson, line 91: For simplicity, how about

```java
ArrayList<String> upCards = (ArrayList<String>) userEntity.getProperty(UPVOTED);
if (upCards == null) {
  upCards = new ArrayList<String>();
}

File: src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java [hunk 7]
```diff
@@ -0,0 +1,174 @@
...
+        downCards.remove(url);
+      }
+    }
+    userEntity.setProperty(UPVOTED, upCards);
+    userEntity.setProperty(DOWNVOTED, downCards);
+    datastore.put(userEntity);
+
+    // Retrieve toggle status of upvote and downvote buttons for user
+    String status = NONE;
+    if (userEntity.getProperty(UPVOTED) != null
```
anbenson, line 130: How about you just look at upCards? It's the same thing, and you don't have to do null checks. (Incidentally, this will never be null since you just set it to be a non-null thing on line 124. Something something proving safety of a null dereference.)

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,174 @@
...
+        && ((ArrayList<String>) userEntity.getProperty(DOWNVOTED)).contains(url)) {
+      status = DOWNVOTED;
+    }
+
+    // Calculate the total votes per card
+    long totalUpvotes = 0;
+    long totalDownvotes = 0;
+    Query allUsers = new Query(USER);
+    PreparedQuery results = datastore.prepare(allUsers);
+    for (Entity entity : results.asIterable()) {
```
anbenson, line 144: There's almost definitely a Datastore way to avoid this loop and do it right from the query (similar to how you can avoid doing a loop by adding a filter predicate), but neither of us have time to look into it. Just know that counting things satisfying a predicate is a basic thing SQL queries can do.

src/main/java/com/google/step/snippet/servlets/FeedbackServlet.java
```diff
@@ -0,0 +1,174 @@
...
+      }
+      if (userDownCards != null && userDownCards.contains(url)) {
+        totalDownvotes++;
+      }
+    }
+
+    // Update card's feedback entity with new vote total
+    Query.FilterPredicate filter = new Query.FilterPredicate(URL, FilterOperator.EQUAL, url);
+    Query query = new Query(FEEDBACK).setFilter(filter);
+    Entity feedbackEntity = datastore.prepare(query).asSingleEntity();
```
anbenson, line 158: The very existence of a Feedback entity seems unnecessary, since you just wrote a bunch of code to determine the number of votes that a piece of feedback has without ever looking at a feedback entity, but again, no time, so okay.

src/main/java/com/google/step/snippet/servlets/SearchServlet.java
anbenson: I'd really prefer to have a shorter timeout. No one waits 30 seconds for a page to load. Isn't this an artifact from when we hypothesized that pages were timing out, but was fixed by adding the background thread factory?

PR #100, opened 2020-07-29
Sanitizing SO client

irisliu77:


src/main/java/com/google/step/snippet/external/StackOverflowClient.java
```diff
@@ -151,7 +153,9 @@ String getResponse(String url, String fieldParam) {
       JSONArray items = json.getJSONArray(ITEM_PARAMETER);
       JSONObject obj = items.getJSONObject(0);
       if (obj.has(fieldParam)) {
-        return obj.get(fieldParam).toString();
+        String result = obj.get(fieldParam).toString();
+        result = Jsoup.clean(result, Whitelist.relaxed());
```
kaylaleung, line 157: can `return Jsoup.clean(result, Whitelist.relaxed());` directly

